PR #1, opened 2026-05-16
docs only: warn about /proc visiblity when `--api-key` is used

hiAndrewQuinn:
Passing an API key via `--api-key` makes it visible to any user on the system through `/proc/<pid>/cmdline`, `htop`, or `ps auxww`.

This PR doesn't change that, it just highlights that in the CLI help text and the logs:

```bash
❯ ./zerostack --help
Minimal coding agent

Usage: zerostack [OPTIONS] [MESSAGE]...

Arguments:
  [MESSAGE]...  Prompt message(s)

Options:
[...]
      --api-key <API_KEY>          API key for the provider (WARNING: visible to other users via ps/htop; prefer env vars)
```

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -27,7 +27,7 @@ pub struct Cli {
     #[arg(long = "model", env = "ZS_MODEL", help = "Model name")]
     pub model: Option<String>,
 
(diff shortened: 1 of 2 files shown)


gi-dellav:
Thanks!

PR #4, opened 2026-05-17
Add max_agent_turns option to control the number of agent turns

dparnell:
I have a long running task that would fail due to the hard coded 100 turn limit. This little patch allows the turn limit to be set via the command line or the config file.


gi-dellav:
Thanks, looks good!

PR #8, opened 2026-05-17
Add Zima Labs as a first-class provider

nikos118:
## Summary

- Adds [Zima Labs](https://zimalabs.io/) as a built-in provider (`--provider zima`), env var `ZIMA_API_KEY`, default base URL `https://zimalabs.io/api/v1`.
- Zima is OpenAI-compatible (chat/completions, Bearer auth), so the new `ProviderKind::Zima` reuses the `openai::CompletionsClient` builder and `AnyClient::OpenAI` — no new variants on `AnyClient` / `AnyModel` / `AnyAgent`.
- The existing `[custom_providers.zima]` override path still works (base URL and api_key_env can be overridden in `config.json`) because `resolve_provider_info` checks custom providers before built-in kinds.

Net change: ~10 lines across `src/provider.rs` and one line in `README.md`.

## Test plan

- [ ] `cargo check` and `cargo build --release` clean
- [ ] `cargo clippy` clean
- [ ] Smoke test with a real key: `ZIMA_API_KEY=zima_sk_... zerostack --provider zima --model <zima-model> -p "hi"`
- [ ] Confirm `Authorization: Bearer <key>` is the correct auth scheme for Zima (per their "works with existing OpenAI SDKs without modification" claim — if their server actually wants the raw header value without `Bearer `, the fix would be local to the Zima arm in `create_client`)
- [ ] Verify override path: a `[custom_providers.zima]` block in `config.json` takes precedence over the built-in defaults

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,7 @@ The worktrees integrations offers 3 slash commands:
 - Anthropic
 - Gemini
 - Ollama
+- Zima Labs
 
 Custom providers can be configured with any base URL and API key environment
 variable in `$XDG_CONFIG_HOME/zerostack/config.json`.
--- a/src/provider.rs
+++ b/src/provider.rs
@@ -19,6 +19,8 @@ use crate::session::SessionMessage;
 #[cfg(feature = "mcp")]
 use crate::extras::mcp::McpClientManager;
 
+const ZIMA_BASE_URL: &str = "https://zimalabs.io/api/v1";
+
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum ProviderKind {
     OpenRouter,
@@ -26,6 +28,7 @@ pub enum ProviderKind {
     Anthropic,
     Gemini,
     Ollama,
+    Zima,
 }
 
 pub fn parse_provider(name: &str) -> Option<ProviderKind> {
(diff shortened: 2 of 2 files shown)


gi-dellav:
Hi! Now that we implemented custom providers (https://github.com/gi-dellav/zerostack/pull/11), i don't think that adding other providers, if they are OpenAI-compatible, is useful for the minimalistic nature of this agent.

However if you want, you can add a guide to set it up using Custom providers in 'providers/ZimaLabs.md', it might be a good idea in order to improve UX with other providers.

nikos118:
Hey @gi-dellav, wanted to add Zima by default, it's one of my fav openrouter alternatives due to its super secure achitecure, and I did not see any privacy first options bundled in by default.

Feel free to read more here but I think it would be a cool edition https://zimalabs.io/

Also - would love to hear your thoughts on adding Composio for tool calls / managed auth :)

lee-b:
Agree, makes no sense to add lots of custom providers that just implement the OpenAI API anyway.

nikos118:
It doesn't just wrap OpenAI. You can give the Zima docs a look. It's a lot more similar to open router, they have a bunch of models that they host them on their own. They don't just wrap them. They have their own GPUs and they host them. They are an inference provider, but they have their own secure architecture where everything's in TEEs based on their own reasearch, and so I don't see any current option that you guys have built in that provides this.

I'm game for a different option. I personally like using Zima, but for any privacy-first deployments, I would love something actually secure where the provider can't see what is being processed.

@gi-dellav

gi-dellav:
Yes, i just mean that the API follows the OpenAI standard, exactly like OpenRouter; I can assure you that if you set up a Custom provider with ZimaLabs's API endpoint, you'll be able to use all of the main advantages offered by Zima.

PR #13, opened 2026-05-17
fix: mutex safety, config validation, /loop startup, and test coverage

gqf2008:
## Summary

  - Replace 14 .lock().unwrap() with poison-safe unwrap_or_else
  - Invalid config JSON now exits with a clear error instead of silent fallback
  - Fix /loop command never starting iteration 1
  - Fix status bar directory display on Windows (Path::file_name())
  - Pass configured max_agent_turns through run_print (was hardcoded 100)
  - Auto-detect shell for loop --run (powershell on Windows, sh elsewhere)
  - Replace let _ = with error reporting for session/transcript saves
  - Add 22 tests for PermissionChecker and EditTool
  - debug = false in release profile

  ## Test plan

  - [x] cargo check — zero errors
  - [x] cargo clippy — no new warnings
  - [x] cargo test — 29/29 passed

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,3 +55,4 @@ opt-level = "z"
 lto = true
 codegen-units = 1
 strip = true
+debug = false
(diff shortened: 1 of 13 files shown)


gi-dellav:
LGTM, thanks!

PR #23, opened 2026-05-17
feat: Add standard rust formatter to CI and run the formatter

paiyar:
Background
---
Add the standard rust formatter `cargo fmt --all --check` to CI and run the formatter

Why?
---
Per https://doc.rust-lang.org/stable/style-guide/ 

> 
> Formatting code is a mostly mechanical task which takes both time and mental effort. By using an automatic formatting tool, a programmer is relieved of this task and can concentrate on more important things.
> 
> Furthermore, by sticking to an established style guide (such as this one), programmers don’t need to formulate ad hoc style rules, nor do they need to debate with other programmers what style rules should be used, saving time, communication overhead, and mental energy.
> 
> Humans comprehend information through pattern matching. By ensuring that all Rust code has similar formatting, less mental effort is required to comprehend a new project, lowering the barrier to entry for new developers.
> 
> Thus, there are productivity benefits to using a formatting tool (such as rustfmt), and even larger benefits by using a community-consistent formatting, typically by using a formatting tool’s default settings.


gi-dellav:
You are right, I am not really experienced with CIs, so i normally just ran it on my machine. Thanks!

PR #53, opened 2026-05-21
feat: OSC 52 clipboard fallback when no external tool available

allen-munsch:
When wl-copy/xclip/pbcopy/clip.exe are all absent, use OSC 52
escape sequences to copy selection to the system clipboard.

Works in most modern terminals (Kitty, Alacritty, WezTerm, foot,
iTerm2, Windows Terminal) without any external dependencies.

Includes a minimal inline base64 encoder with tests.

### DIFF
--- a/src/ui/renderer.rs
+++ b/src/ui/renderer.rs
@@ -685,4 +685,29 @@ pub fn copy_to_clipboard(text: &str) {
             return;
         }
     }
+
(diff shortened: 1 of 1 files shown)


gi-dellav:
Will merge for v1.3.0 (later today), thanks

allen-munsch:
Thank you for sharing zerostack, wonderful project!

PR #57, opened 2026-05-21
fix(input): discard whitespace-only input instead of submitting it

### LINKED ISSUE #56: Whitespace-only input is submitted to the model (provider-dependent failure)
## Summary
When the input buffer contains only whitespace (spaces, tabs, or newlines from Shift/Alt+Enter), pressing Enter submits it to the provider as a user message. The result depends on which provider is selected:
- **Anthropic** (e.g. `claude-haiku-4-5`) rejects the request with `400 invalid_request_error: messages: text content blocks must contain
    non-whitespace text`, which surfaces as an error in the TUI.
- **OpenAI** accepts the request, so the message is sent and counted

Either way, this is almost never the user's intent — it usually happens from an accidental space + Enter — and the behaviour is inconsistent across providers.

## Steps to reproduce
1. Launch `zerostack`.
2. 2. At the `>` prompt, press the space bar one or more times (no other characters).
3. Press Enter.

## Expected
The whitespace input is discarded. The prompt returns to its idle state, the same way it does when Enter is pressed on a truly empty buffer. This should be consistent regardless of provider.

## Actual
With Anthropic: provider returns a 400 error (screenshot below).

Image

## Environment
- zerostack: current `main` (8c862e3)
- Reproduced on Anthropic (`claude-haiku-4-5`); OpenAI accepts but exhibits the same root cause.
- OS: macOS

## Proposed fix
Treat whitespace-only input as empty in `InputEditor::handle_key` (`src/ui/input.rs`): switch the submit guard from `text.is_empty()` to `text.trim().is_empty()`. 
This unifies behaviour across providers and prevents accidental submissions regardless of how the provider would have handled them.

xavierforge:
## Summary
- Treat a buffer containing only whitespace as empty in `InputEditor::handle_key`, so pressing Enter on it clears the prompt instead of sending a blank message.
- Unifies behaviour across providers: Anthropic previously rejected the request with a 400 `messages: text content blocks must contain non-whitespace text`; OpenAI silently accepted the empty message.
- Whitespace-only input is also no longer pushed onto the input history.

Fixes #56

### DIFF
--- a/src/ui/input.rs
+++ b/src/ui/input.rs
@@ -265,13 +265,14 @@ impl InputEditor {
                     return None;
                 }
                 let text = self.buffer.clone();
-                if !text.is_empty() {
+                let is_blank = text.trim().is_empty();
+                if !is_blank {
                     self.history.push(text.clone());
                 }
                 self.history_pos = None;
                 self.buffer.clear();
                 self.cursor = 0;
-                if text.is_empty() { None } else { Some(text) }
+                if is_blank { None } else { Some(text) }
             }
             KeyCode::Char(c)
                 if c == '\x08' || (c == 'h' && key.modifiers.contains(KeyModifiers::CONTROL)) =>


gi-dellav:
Will merge for v1.3.0 (later today), thanks

xavierforge:
Thanks, and also thanks for bringing this project. 
Love it.

PR #59, opened 2026-05-21
Authentication/Key Resolution Refactoring

skorotkiewicz:
## Overview
Refactor the authentication/key resolution system to eliminate code duplication, consolidate the priority chain (CLI args → env vars → config), and simplify provider client building.

### Create Unified Auth Resolution Module
- Created `src/auth.rs` with `AuthResolver` struct and `ProviderKind` enum
- Implemented builder pattern for configuring key resolution sources
- Simplified key resolution logic with clear priority chain
- Added comprehensive tests for all resolution scenarios

### Simplify Provider Info Resolution
- Renamed `ProviderInfo` to `ProviderConfig` with cleaner API
- Changed return type from `Option` to `Result` for better error handling
- Removed redundant `CUSTOM_BASE_URL` env var fallback (now handled via custom_providers config)

### Consolidate Client Building with Factory Pattern
- Extracted client builder functions into separate `build_*_client` functions
- Removed `Custom` provider variant (now an alias for `openai` with custom base_url)
- Simplified `create_client` to use `AuthResolver` for key resolution

### Simplify `build_agent` (PARTIALLY COMPLETED)
- Client building consolidated
- Agent building remains similar due to different generic types per client

### Remove Custom Provider
- `custom` provider type now maps to `openai` client with custom base_url
- Removed `AnyClient::Custom`, `AnyModel::Custom`, `AnyAgent::Custom` variants
- Updated tests to reflect new behavior

### Custom Provider Name Key Lookup
- Added `with_custom_provider_name` to `AuthResolver`
- Config keys under `api_keys` can now use custom provider names (e.g., `"local-vllm"`) as well as provider slugs (`"openai"`)
- This allows configs like:
  ```json
  {
  "model": "gpt-uncensored",
  "provider": "local-llm",
    "custom_providers": {
      "local-llm": {
        "provider_type": "openai",
        "base_url": "http://localhost:8642/v1"
      }
    },
    "api_keys": {
      "local-llm": "sk-..."
    }
  }
  ```

---

### `src/auth.rs` (new file)
- `ProviderKind` enum with `from_name()` method
- `AuthResolver` struct with builder pattern for configuration
- Clean `resolve()` method with priority: CLI key → env var → config file

### `src/provider.rs` (modified)
- Removed `ProviderKind` definition (moved to `auth.rs`)
- Removed `resolve_api_key` function (replaced by `AuthResolver`)
- Renamed `ProviderInfo` to `ProviderConfig`
- Simplified `resolve_base_url` to just return `config.base_url.clone()`
- Removed `Custom` variants from `AnyClient`, `AnyModel`, `AnyAgent` enums
- Added `build_*_client` helper functions

### `src/tests/auth_tests.rs` (new file)
- 11 tests covering all `AuthResolver` scenarios
- Tests for `ProviderKind::from_name` behavior
- Isolated test environment variable modifications


allen-munsch (commented):
Nice refactor — the builder pattern on `AuthResolver` is clean and the priority chain is much clearer than the old scattered resolution.

One suggestion on the test approach: the `unsafe { std::env::set_var(...) }` + `ENV_LOCK` pattern has a subtle issue. `set_var` mutates process-global state, and `cargo test` runs tests in parallel across threads by default. Even with the lock serializing within this module, any other test anywhere in the crate that reads `OPENAI_API_KEY` can see the wrong value mid-flight — `std::env::set_var` is `unsafe` specifically because it can't be made thread-safe.

I ran into the same thing, and was told about a closure-based approach — instead of calling `std::env::var` directly inside `resolve()`, accept the env lookup as a parameter:

```rust
// Inject the env lookup so tests can supply a mock.
pub fn resolve_with_env<F: Fn(&str) -> Result<String, VarError>>(
    &self,
    get_env: F,
) -> anyhow::Result<String> {
    // ... use get_env(env_var) instead of std::env::var(env_var)
}

// Convenience for production callers.
pub fn resolve(&self) -> anyhow::Result<String> {
    self.resolve_with_env(|name| std::env::var(name))
}
```

Then tests pass a `HashMap`-backed closure — zero `unsafe`, zero `set_var`, no lock needed, fully parallel-safe:

```rust
fn mock_env(vars: &[(&str, &str)]) -> impl Fn(&str) -> Result<String, VarError> {
    let map: HashMap<String, String> = vars
        .iter()
        .map(|(k, v)| ((*k).to_string(), (*v).to_string()))
        .collect();
    move |name: &str| map.get(name).cloned().ok_or(std::env::VarError::NotPresent)
}

#[test]
fn auth_resolver_returns_env_var_when_no_cli_key() {
    let env = mock_env(&[("OPENAI_API_KEY", "env-key-456")]);
    let resolver = AuthResolver::new(ProviderKind::OpenAI)
        .with_cli_key(None)
        .with_config_keys(None);
    assert_eq!(resolver.resolve_with_env(env).unwrap(), "env-key-456");
}
```

Same reasoning as your builder pattern — push the side effect to the boundary, keep the core pure.

gi-dellav:
Thanks; i just release in v1.3.0, but tomorrow I will take the time to look in detail to this PR together with #64 and I will decide on how to design a better auth system for v1.3.1 (hopefully released in ~24 hours).

Cheers,
G.

gi-dellav:
Just merged, check out v1.3.1 and see if everything works fine

PR #60, opened 2026-05-22
feat: bracketed paste, Emacs keybindings, and paste event

sqwxl:
### Summary

Three logical changes split into separate commits:

1. **Terminal** — enable bracketed paste and keyboard enhancement flags on enter, disable on teardown
2. **Input** — Emacs-style keybindings; standard for most terminal text inputs. (C-a, C-e, C-f, C-b, C-d, C-k, C-y, M-d, M-Backspace, C-w), kill ring, yanking, multi-line navigation, history draft preservation
3. **Event system** — `Paste(String)` event variant wired through the event loop to `InputEditor::handle_paste()`


gi-dellav:
Hey, 

considering the minimalistic design of zerostack, shouldn't Ctrl+G (aka open in external editor) be enough, or do you think that we should implement those features inside our TUI renderer?

Cheers,
G.

sqwxl:
hey thanks for making this, i appreciate the minimalism a lot! and i certainly don't want to introduce unnecessary bloat.

for the keyboard shortcuts, while it's true dropping into an external editor could be enough, i personally am very used to the standard cursor and editing shortcuts added here. they are very common across most shells (bash, zsh, fish) and many terminal apps that accept text input. i think many users would expect these to work here, too; and most other coding agents i've tried support these to various degrees (claude, codex, opencode)

for the paste enhancement (commits 1 and 3), i think those are net positives; they prevent an issue i noticed when pasting in multiline text where the input would interpret a newline in the pasted text as a literal 'return' and submit the preceeding fragment to the model.

gi-dellav:
I thought about it for some time, and I actually think that some extra shortcuts and QoL for the TUI, and +900 LoC is reasonable considering that we are shipping a custom renderer.

So, thank you for your contribution!

(P.s. will be merged later today)

Cheers,
G.

PR #61, opened 2026-05-22
fix: eliminate cursor and input area flicker during rendering

sqwxl:
Fixes two sources of flicker:

1. Cursor flicker on empty lines - terminal hardware cursor was visible during render_viewport as it moved through column 0 of every row. Now hidden via crossterm cursor Hide during rendering and only shown after the cursor reaches its correct input-area position.

2. Input area and status line flicker during typing - every keystroke triggered a full render_viewport (chat area redraw) even though only the input buffer changed. Now skips the viewport for input-only edits when agent is idle. Falls back to full redraw when agent is streaming (is_running) to ensure latest tokens remain visible.

### DIFF
--- a/src/ui/mod.rs
+++ b/src/ui/mod.rs
@@ -588,8 +588,16 @@ pub async fn run_interactive(
                                     );
                                 }
                             }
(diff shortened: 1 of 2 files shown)


gi-dellav:
LGTM, thanks!

PR #62, opened 2026-05-22
Add danger_accept_invalid_certs toggle for custom providers

s3rj1k:
Allows connecting to local inference providers that use selfsigned certs without messing around with CA storage


gi-dellav:
LGTM, thanks!

PR #64, opened 2026-05-22
feat(provider): selectable OpenAI API style + custom headers

### LINKED ISSUE #63: GPT-5 models fail with 400 "max_tokens not supported" on the OpenAI provider
## Summary
Using any GPT-5-series model through the `openai` provider fails with a 400 from the OpenAI API. 
zerostack's OpenAI client is built on rig's `openai::CompletionsClient` (the Chat Completions transport), which sends the legacy `max_tokens` parameter. GPT-5 models reject `max_tokens` on `/chat/completions` and require `max_completion_tokens` instead.

## Steps to reproduce
1. Set `OPENAI_API_KEY`.
2. Run: `zerostack --provider openai --model gpt-5 -p "hi"`

## Actual behavior
Error: ... 400 ... "Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead." ...
Image

## Root cause
rig 0.37 exposes two OpenAI client types:
- `openai::Client` → Responses API (`/responses`); maps `max_tokens` to `max_output_tokens`, which GPT-5 accepts.
- `openai::CompletionsClient` → Chat Completions API (`/chat/completions`); sends `max_tokens` verbatim.

zerostack wires the OpenAI (and custom) provider to `CompletionsClient`, so all GPT-5 traffic hits the rejected parameter. 
Note the two endpoints are not interchangeable: many OpenAI-compatible gateways implement only `/chat/completions` and have no `/responses`, so blindly switching everything to the Responses client would instead break those gateways with a 404.

### Proposed fix
Make the API style selectable so real OpenAI can use Responses (fixing GPT-5) while compatible gateways keep using Chat Completions. PR to follow.

### Environment
- zerostack: current main ([8c862e3](https://github.com/gi-dellav/zerostack/commit/8c862e3e7e904c3d776e20f2c1d19cee52552eda))
- rig: 0.37

xavierforge:
Fixes #63 

## What
Adds an `api_style` setting to the OpenAI provider so it can use either rig's Responses API client or its Chat Completions client, and adds support for custom HTTP headers on OpenAI-compatible providers.

## Why
- **GPT-5 was broken.** zerostack used rig's `CompletionsClient` for all OpenAI traffic, which sends `max_tokens` — rejected by GPT-5 (`Use max_completion_tokens instead`, 400). The Responses client maps `max_tokens` to `max_output_tokens` and works.
- **But Responses isn't universal.** Many OpenAI-compatible gateways implement only `/chat/completions`; forcing Responses would 404 against them. The two transports need to coexist, not be swapped wholesale.
- **Self-hosted gateways often need extra headers** (e.g. Cloudflare Access), which there was previously no way to set.

## How
- Introduce `OpenAiClient` / `OpenAiModel` / `OpenAiAgent` inner enums holding either the Responses or Completions variant.
- Add `ApiStyle` config. Default: Responses for plain `openai` (no base_url), Completions when a custom `base_url` is set; overridable per provider.
- Add a `headers` map (with `${ENV_VAR}` expansion), plus opt-in `insecure_skip_verify` and `timeout_secs`, wired through a custom `reqwest::Client`.

## Testing
- Real OpenAI + `gpt-5.x` → Responses path, no more 400. ✅
- Self-hosted gateway behind Cloudflare Access → Completions + CF headers →
  success. ✅
- Same gateway with a deliberately wrong CF secret → 403 from Cloudflare,
  confirming headers are actually sent. ✅
- Unit tests for `resolve_api_style` and `expand_env`.

## Notes / coordination
This touches `src/provider.rs`, which overlaps with #59  (auth refactor + removal of the `Custom` variant). 
I intentionally did **not** remove `Custom` here to avoid duplicating that PR's change. 
Happy to rebase whichever lands first.


gi-dellav:
Why does it add more than 300 LoC for just a new provider support (aka OpenAi client compatible with 5.x family models)?

It seems like the PR is touching other parts of the codebase, not just OpenAI support

xavierforge:
Good question. let me break down where the size comes from.

The PR actually covers two related things for OpenAI / OpenAI-compatible
providers:
1. **GPT-5 support (the main goal).** rig 0.37 exposes two *different* OpenAI
   client types — `openai::Client` (Responses API) and
   `openai::CompletionsClient` (Chat Completions). zerostack used the latter,
   which sends `max_tokens` and gets a 400 from GPT-5 (`Use
   max_completion_tokens`). Supporting both isn't a one-liner: because the two
   produce different model types (`ResponsesCompletionModel` vs
   `completion::CompletionModel`), the `AnyClient` / `AnyModel` / `AnyAgent`
   enums each need an inner enum to hold either variant, which is what fans out
   the match arms in `completion_model`, `summarize_with_model`, `run_print`,
   `spawn_runner` and `build_agent`. Most of the added lines are this
   type-driven boilerplate, not new logic.

2. **Custom HTTP headers** for OpenAI-compatible gateways behind auth proxies
   (e.g. Cloudflare Access), with `${ENV_VAR}` expansion so secrets stay out of
   config. The actual new logic here is small (`build_http_client`,
   `expand_env`, `resolve_api_style` — under ~50 LoC combined).

You're right that (2) is a separate concern. They're bundled because both touch
the same OpenAI client construction path, but I'm happy to split this into two
PRs (api_style first, headers second) if you'd prefer to review them
independently — just let me know.

Also heads up: I'm currently rebasing onto #62, since its
`danger_accept_invalid_certs` change overlaps with this. I've dropped my own
TLS toggle in favor of #62's and folded headers/timeout into the shared
`http_client` it introduced. Pushing the resolved version shortly.

gi-dellav:
Am I talking to an AI?

xavierforge:
Sorry that I gave you this feeling.
English isn't my first language, so I use AI to help me translate and phrase things. 
That's probably why my comments read that way.

But the change itself is mine and I understand it. 
Short version in my own words: rig 0.37 has two OpenAI clients (the Responses one and the Completions one), and they're different Rust types, so I had to wrap them in an enum to support both. 
This project used the Completions one, which is why GPT-5 fails (it sends `max_tokens`). 
I hit this with my company's gateway, which sits behind Cloudflare Access, so I also needed custom headers, that's the second part.

I tested both: GPT-5 on real OpenAI now works, and my gateway works with the CF headers (and correctly returns 403 when I set a wrong header, so I know the headers are actually sent).

Happy to walk through any part of the diff, or split it into two PRs if that's easier to review.

xavierforge:
BTW, on the line count.
I went back and checked the diff. It's +397 across 3 files, but ~327 of those lines are in `provider.rs`, and almost all of them are type wrapping. 
The two rig clients are different types, so `AnyClient`/`AnyModel`/`AnyAgent` each needs an inner enum, and every match on them grows extra arms. 
The rest is small: CONFIG.md is ~55 lines (the new docs section plus the config example), and config/mod.rs is ~15 (the `ApiStyle` enum and the new fields). 
So the size is mostly the enum boilerplate, which bloats the LoC.

gi-dellav:
Thanks, now I looked a little bit at the design and it makes sense.
I just released v1.3.0, tomorrow I will take a deep look at this PR and at #59 , so that v1.3.1 (hopefully out in ~24hours) will have a fixed auth system with full OpenAI support.

Thx,
G.

gi-dellav:
Just merged, check out v1.3.1 and see if everything works fine

xavierforge:
Thanks. All is working fine.
And I also notice the start-up time seems faster😎

gi-dellav:
Yes, in v1.3.0 I finished implementing lazy loading of some objects like MCPs, so that they are loaded only when the user executes a prompt.

Anyways, thanks for your contribution and happy coding with zerostack!

PR #68, opened 2026-05-25
Fixed the issue where the cursor’s position during input of CJK chara…

gaoshang96:
Fixed the issue where the cursor position during input of CJK characters (Chinese, Japanese, Korean characters) was incorrect. This issue involved restructuring the logic related to character width and cursor positioning in files like input/mod.rs, pickers.rs, markdown.rs, renderer.rs, and utils.rs.


gi-dellav:
LGTM, will merge later today, thanks for the contribution

PR #70, opened 2026-05-26
Add an optional plain-Markdown persistent memory feature

xavierforge:
## Summary

Adds an optional, plain-Markdown persistent memory system, gated behind a new `memory` feature (off by default, in the spirit of `acp`). 
When enabled, zerostack remembers preferences and recent context across sessions by keeping Markdown files
on disk and injecting the relevant ones into the system prompt (no embeddings, no
database, just files the user can read, edit, and commit.).

## What it does
- **Four kinds of memory** under `<data_dir>/agent/memory/`:
  - `MEMORY.md` — curated long-term memory. **Global**, always injected.
  - `projects/<slug>/SCRATCHPAD.md` — per-project checklist; only open `- [ ]` items injected.
  - `projects/<slug>/daily/YYYY-MM-DD.md` — per-project log; today + yesterday injected.
  - `projects/<slug>/notes/*.md` — per-project reference, on-demand only.
- **Per-project scoping**: `<slug>` derives from the cwd, so unrelated projects don't leak context into each other; only `MEMORY.md` is shared globally.
- **Three rig tools**: `memory_write`, `memory_read`, `memory_search`. Search is multi-term, case-insensitive, returns hits with surrounding context, ranks by relevance, and renders within a byte budget; it locates files for `memory_read`
  to read in full.
- **Compaction-safe**: each compaction summary is deterministically flushed to the daily log *before* `Session::compress`, so context survives compaction instead of depending on the model to persist it.
- **Budget-aware**: the injected block lives in the preamble, which `needs_compaction()` doesn't count, so its estimated tokens are folded into the reserve to trigger compaction early enough.

## Design notes
- **Off by default.** Everything is behind `#[cfg(feature = "memory")]`; default builds are completely unaffected. Whether to include it in `default` is left to your judgement.
- **No new heavy deps.** Only `regex` (for search); project slugs use std `DefaultHasher`. No embeddings/vector store, keeping the footprint small.
- **Snapshot injection**: memory is read into `ContextFiles` at load/reload, so long-term memory written mid-session takes effect from the next session. This is intentional (keeps I/O off the hot path); documented in the README.
- **Known limitation**: search is literal/line-based, so it misses cross-line phrasings and synonyms; `estimate_tokens` (chars/4) underestimates CJK. Both are noted with TODOs for a future semantic-retrieval / better-tokenizer pass.

## Possible follow-ups (left for maintainer)
- A `--no-memory` runtime opt-out (when compiled in but you want a clean run).
- Considering `memory` for the `default` feature set.

## Commits
Split by capability, each compiles and tests independently:

1. `feat(memory)`: core store + tools (single-term search)
2. `feat(memory)`: per-project scoping
3. `feat(memory)`: multi-term ranked search with a render budget
4. `feat(memory)`: inject into the system prompt + register tools
5. `feat(memory)`: persist compaction summaries to the daily log
6. `fix(memory)`: account for injected memory in the compaction budget
7. `docs(readme)`: document the feature

## Testing
`cargo test --features memory` covers the store (write/append/overwrite, scratchpad filtering, daily ordering, path-traversal rejection, CJK-safe truncation), search (context windows, merging, block cap, multi-term ranking, MEMORY.md priority, render summary/budget), per-project isolation, preamble injection, compaction flush, and the reserve calculation.

Default (non-memory) build, and tests also pass unchanged.


gi-dellav:
Looks good, will check out later today and then merge in v1.4.0 (1/2 weeks)

PR #73, opened 2026-05-30
Add model field to custom providers for default model selection

skorotkiewicz:
Custom providers can now specify a default model that overrides the system
default when no explicit model is set via CLI or environment variable.


gi-dellav:
Thanks for your contribution!

P.S. I see that you are a recurrent contributor, you might be interested in joining the Element's chatroom in order to keep up with the developments ( https://matrix.to/#/#zerostack-general:matrix.org )

Cheers,
G.

PR #82, opened 2026-06-02
fix: default the model to the chosen provider (startup and /provider)

xavierforge:
## Demo
https://github.com/user-attachments/assets/7435fd4a-85de-4dd6-9015-0b57be8b64e1

## Summary
Choosing a provider without also choosing a model left the model at the OpenRouter default id (`deepseek/deepseek-v4-flash`), which is invalid on every other provider. 
This happened in two places:
- **At startup**: `zerostack --provider anthropic` (or `provider` in config) with no `--model` kept the deepseek id, and the header banner showed it.
- **At runtime**: `/provider anthropic` kept the previous model id.

Both now derive a sensible default for the target provider, with this priority:
1. a custom gateway's configured `model`,
2. a quick model targeting that provider (carrying its pricing),
3. a built-in fallback.

Built-in fallbacks: anthropic to `claude-sonnet-4-6`, openai to `gpt-5.1`, gemini to `gemini-2.5-pro`, openrouter to `openrouter/auto`, ollama to `llama3.1`. 
A user's quick model or a custom gateway model for the provider always wins over these.

When no provider is chosen explicitly, the historic default (`deepseek/deepseek-v4-flash` on openrouter) is preserved unchanged.

## How it works
A single `default_model_for_provider` helper in `provider.rs` is the source of truth. 
`resolve_model` uses it so the startup banner, the status line, and the session all agree, and `handle_provider` uses it so a runtime switch reports the resulting model in its confirmation message.

## Commits
1. fix(provider): default to a sensible model when switching providers
2. fix(cli): default the model to the chosen provider at startup

## Testing
- `cargo build` clean (no warnings).
- `zerostack --provider anthropic` with no `--model` starts on `claude-sonnet-4-6` (header and status line agree); plain `zerostack` still starts on `deepseek/deepseek-v4-flash`.
- `/provider anthropic` at runtime switches the model and reports it.

## Notes
- Independent of the live-models / pickers PR. They share `provider.rs` and `slash/providers.rs` but edit different functions; a trial merge is clean, and the merged tree builds and passes tests, so the two can land in any order.
- Built-in fallback IDs drift with provider releases; adjust to taste. User quick models and custom gateway models take priority over them.
- Known limitation: the header banner reflects startup resolution, so after a runtime `/provider` switch, the banner keeps the startup model while the status line updates. This is pre-existing and left for a follow-up.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -200,16 +200,22 @@ impl Cli {
     }
 
     pub fn resolve_model(&self, cfg: &config::Config) -> CompactString {
-        self.model
-            .as_deref()
-            .or(cfg.model.as_deref())
-            .map(CompactString::new)
-            .unwrap_or_else(|| {
-                let qm = config::quick_models_map(cfg);
-                qm.get("deepseek-v4-flash")
-                    .map(|q| q.model.clone())
-                    .unwrap_or_else(|| CompactString::new("deepseek/deepseek-v4-flash"))
-            })
+        if let Some(m) = self.model.as_deref().or(cfg.model.as_deref()) {
+            return CompactString::new(m);
+        }
+        // No explicit model. If a provider was chosen explicitly, default to a
+        // model valid for it so `--provider anthropic` does not keep the
+        // OpenRouter default id; otherwise keep the historic deepseek default.
+        if (self.provider.is_some() || cfg.provider.is_some())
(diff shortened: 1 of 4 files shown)


gi-dellav:
Thanks, merged!

PR #94, opened 2026-06-04
Convert compaction summary to Assistant role for strict chat templates

### LINKED ISSUE #93: /compact errors with "System message must be at the beginning" on Qwen 3.x
Running Zerostack against a local llama.cpp serving Qwen 3.6 35B-A3B MTP (same chat-template family as Qwen 3 generally). Up through the first few thousand tokens of a session everything works. Then I run `/compact` and the next agent request fails:

      error: http client error: ... Invalid status code 400 ...
      {"error":{"message":"... Jinja Exception:
      System message must be at the beginning.","type":"invalid_request_error"}}

The agent gives up mid-turn and drops back to the prompt. It's deterministic. Every time `/compact` fires against this model, the next request blows up the same way.

What's happening: llama.cpp itself accepts the request fine. The rejection comes from Qwen's Jinja chat template, which explicitly refuses any System message past position 0. I checked the template; the constraint is intentional, not a llama.cpp policy thing.

On the Zerostack side I traced it to `convert_history` in `src/agent/runner.rs`, which emits the prior conversation summary as `Message::system`. Since the agent preamble is already a System message, the outgoing request ends up with two System messages stacked at the start of `messages[]`. Qwen's template raises on the second one.

I have a small fix that's been clean in my testing.

A few notes that might help:
1. This isn't something my fork introduced. It's been latent. It
    triggers any time the existing `/compact` command actually
    fires against a Qwen 3.x model. So if you've never personally hit
    compaction on a Qwen session, you wouldn't have seen it.
2. I suspect other strict templates have similar constraints
    (Mistral's, some Llama 3.x variants), but I haven't actually
    verified those.

hal-eisen:
Closes #93.

Quick diagnosis: in `convert_history` (`src/agent/runner.rs`), the compaction summary was being pushed as `Message::system`. Since the agent preamble is already a System message, requests ended up with two System messages stacked at positions 0 and 1. Qwen 3.x's chat template raises on the second one.

Fix: emit the summary as `Message::assistant`, framed with a `[Recap of my prior work in this conversation]` prefix so the agent reads the content as a self-recap rather than as a fresh turn it's about to continue word-for-word. Assistant role has two advantages over the obvious alternative of User:

- It produces clean User↔Assistant alternation when the next user prompt arrives, matching the standard resumed-conversation shape every chat template expects.
- It reads naturally as "the agent recaps what it did, then the user continues," which is closer to the actual semantic of compaction.

Any persisted `MessageRole::System` entries elsewhere in `session.messages` (a defense-in-depth case — `add_message` doesn't currently produce them, but old sessions or future code paths might) get the same Assistant treatment for code-symmetry with the summary push.

Zerostack's actual system prompt still goes through the rig preamble, which is correctly at position 0. So a typical request now looks like `[System(preamble), Assistant(summary), ...kept, User(latest)]` — one System at position 0, exactly what Qwen's template expects, and clean alternation through the rest.

Tested locally against Qwen 3.6 35B-A3B: `/compact` succeeds and the conversation continues. Without the change, the Jinja exception in #93 reproduces every time.

I'd expect this to also help other strict-template models with the "System at position 0 only" convention — Mistral and some Llama 3.x variants come to mind — but I haven't verified those.

### DIFF
--- a/src/agent/runner.rs
+++ b/src/agent/runner.rs
@@ -98,9 +98,20 @@ pub fn convert_history(session: &Session) -> Vec<Message> {
     let extra = if summary.is_some() { 1 } else { 0 };
     let mut messages = Vec::with_capacity(remaining + extra);
 
+    // The compaction summary is emitted as an Assistant message rather
+    // than a System message: the agent already has a System preamble
+    // (SYSTEM_PROMPT + mode prompt + context files), and some model chat
+    // templates (notably Qwen 3.x) refuse any System message past
+    // position 0. Assistant role also produces clean User↔Assistant
+    // alternation when the next user prompt arrives, which reads as
+    // "the agent recaps what it did, then the user continues" — a
+    // natural resumed-conversation shape. The "[Recap of my prior work
+    // in this conversation]" prefix labels the message as a self-recap
+    // so the agent doesn't treat it as a fresh continuation of its own
+    // voice.
     if let Some(summary) = summary {
-        messages.push(Message::system(format!(
-            "[Previous conversation summary]\n{}",
+        messages.push(Message::assistant(format!(
+            "[Recap of my prior work in this conversation]\n{}",
             summary
         )));
     }
@@ -109,7 +120,12 @@ pub fn convert_history(session: &Session) -> Vec<Message> {
         match msg.role {
             MessageRole::User => messages.push(Message::user(msg.content.to_string())),
             MessageRole::Assistant => messages.push(Message::assistant(msg.content.to_string())),
-            MessageRole::System => messages.push(Message::system(msg.content.to_string())),
+            // Convert any persisted System messages to Assistant for the
+            // same reason as the summary above: the templates that reject
+            // mid-stream System tolerate Assistant, and code-symmetry with
+            // the summary push keeps the resumed-conversation shape
+            // consistent.
+            MessageRole::System => messages.push(Message::assistant(msg.content.to_string())),
         }
     }
 


gi-dellav:
Thank you for the contribution. 
Before merging, I wanted to ask: why does the new compaction message act as User and as Assistant? If it was Assistant, it would have been just a continuation of the agent flow

hal-eisen:
Good question. I went back and forth on this. I can see now that Assistant is the cleaner fit.

My original reasoning for User was: the `[Previous conversation summary]` prefix framed the content as *meta-information about prior turns* rather than the agent's own output, so it felt User-shaped (system-provided context the agent reads as input). Either sidesteps the Qwen Jinja constraint.

But the conversation-flow argument is stronger. With User, the sequence after `/compact` is `User(summary) → User(next prompt)` (two consecutive User messages). Some templates tolerate this, some don't, and even when allowed the model has to figure out which is the "real" user input. With Assistant, you get `Assistant(summary) → User(next prompt)` for a clean alternation, the standard resumed-conversation shape.

The "but the summary wasn't actually written by the agent" objection is a minor fiction either way (it's typically the summarizer model, not the main agent). Both User and Assistant are conventional fictions about role labels; conversational flow weighs more than literal truth.

Pushed the change. The summary now goes as `Message::assistant` with a `[Recap of my prior work in this conversation]` prefix to signal it's a self-recap rather than a fresh continuation. Persisted-System fallback also moved to Assistant for code-symmetry. Updated commit message and PR description to match.

Got me thinking. User/Assistant/System probably doesn't cleanly capture "agent-state metadata like a compaction summary". There's an argument for a fourth role to handle that case properly. Definitely a separate discussion if it's worth pursuing.

gi-dellav:
Thanks, checked and merged

PR #110, opened 2026-06-14
Opt-in mid-turn compaction with a hard ceiling

### LINKED ISSUE #109: Mid-turn context blowout: compaction never fires within a single long turn
## Problem

zerostack only compacts the conversation *between* turns (on `AgentEvent::Done`), against the session's accumulated text-message token estimate. A single long turn (many tool calls and large tool results) can drive the model's real context past its limit *before* that check ever runs, because the in-flight tool traffic lives only in the runner and never enters the session's token estimate. On a fixed-size local KV cache (e.g. llama.cpp at 32k) this reliably overruns the window mid-turn, causing the server to error or silently set `truncated = 1`, which corrupts the conversation.

Phase 1 (commit a32a622) already wired real provider-reported usage via rig's `CompletionCall`, so the pressure is now observable mid-turn. What's missing is acting on it.

## Proposal

Add an **opt-in** within-turn compaction lever, gated by the existing `compact_enabled` master switch:

- New config key `mid_turn_compact_threshold` (fraction of the context window, e.g. `0.80`). **Unset by default**, so out-of-the-box behavior is unchanged.
- When set and `compact_enabled` is true, compact during a turn once real provider prompt pressure crosses the threshold: abort at a clean `CompletionCall` boundary (before the just-returned tool calls execute, so no half-applied edits), record a progress recap, compact, and respawn on the compacted history with a continuation prompt.
- **Runaway guard:** if the next provider call after a compaction is still over the ceiling, compacting again is futile (the irreducible floor — system prompt, tool schemas, kept-recent transcript, reserved response — exceeds the budget). Stop the turn and print the full arithmetic plus concrete options, rather than looping forever.

Single hard threshold only (no soft limit), and the current summarization technique is kept as-is. Motivation for opting in at all also includes avoiding context rot in large windows.

PR follows.

hal-eisen:
Closes #109.

Adds an opt-in lever to compact the conversation *during* a turn, not just between turns, so a single long turn can't silently overrun a fixed context window (e.g. llama.cpp's KV cache).

## Commits

1. **Wire mid-turn token observability via rig `CompletionCall`** — forwards real provider-reported usage mid-stream and uses it as a floor for the status-bar estimate (observability only; no behavior change).
2. **Add opt-in mid-turn compaction with a hard ceiling** — the trigger that acts on it.

## Behavior

- New config key `mid_turn_compact_threshold` (fraction of the context window). **Unset by default → no behavior change.** Gated by the existing `compact_enabled` master switch.
- When set and enabled: on an over-threshold `CompletionCall`, abort the run at that clean boundary (the model's just-returned tool calls haven't executed, so no half-applied edits), record a turn-progress recap, compact, and respawn on the compacted history with a hardcoded continuation prompt. The dominant relief is dropping the aborted run's in-flight tool context, which never enters the session estimate.
- **Runaway guard:** if the next provider call after a compaction is still over the ceiling, compacting again is futile, so the turn is stopped with the full arithmetic (window, ceiling, post-compaction prompt, overflow, reserve, kept-recent) and concrete options. A relief flag distinguishes "compaction failed" from later fresh growth, so a healthy long turn can still compact again.
- Out-of-range threshold values are ignored with a one-time startup warning (the resolver runs per call and must not log).

## Design notes

Single hard threshold (no soft limit); current summarization technique unchanged. Full rationale and the implementation decisions (const continuation prompt vs. a `prompts/*.md` mode file, the abort boundary, continuity caveats) are in `docs/COMPACTION_PLAN.md`.

## Testing

`cargo test` (278 pass), `cargo build` and `cargo clippy` clean (no new warnings). Added unit tests for threshold validation and the exhaustion-report arithmetic. The hard-stop guard's wiring is unit-tested at the arithmetic level; tripping it end-to-end needs a genuinely too-small `context_window` and hasn't been exercised against a live server yet.

### DIFF
--- a/docs/COMPACTION_PLAN.md
+++ b/docs/COMPACTION_PLAN.md
@@ -1,0 +1,177 @@
+# Mid-Turn Compaction — Design Notes
+
+Status: implemented on branch `pr/mid-turn-compaction`. Phase 1 (observability) in commit `a32a622`; phase 2 (the config-gated trigger) follows. Captured 2026-06-01, revised 2026-06-14 after maintainer (gi-dellav) feedback on Matrix.
+
+## Revision log
+
+**2026-06-14 (gi-dellav, Matrix):**
+
+- **Drop the soft limit.** Two thresholds (soft plus hard) are too complex. Keep a single hard limit only.
+- **Make the hard limit a config setting, default off.** If the user does not set a value, mid-turn compaction does not run at all; the pre-existing between-turn behavior is unchanged. Setting a value opts in: compaction fires when real mid-turn prompt pressure crosses that percentage. Docs recommend **80%** as a reasonable starting value.
+- **Keep the current summarization technique;** it is fine as is. The motivation for opting in at all is avoiding context rot in long contexts (see <https://garrit.xyz/posts/2026-05-06-dont-trust-large-context-windows>); a tighter, user-chosen ceiling keeps the live prompt small enough that the model still attends to it.
+
+**2026-06-14 (implementation):**
+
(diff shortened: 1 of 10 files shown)


gi-dellav:
Sorry for the delay, will check in the next day

Cheers

PR #116, opened 2026-06-19
fix: resolv.conf may be a symlink

### LINKED ISSUE #115: bwrap cannot bind resolv.conf when it is a symlink to a separate filesystem
**Describe the bug**
When `resolv.conf` is a symlink whose target lives on a filesystem that is not covered by the sandbox's --ro-bind / / mount, the hard-coded `--ro-bind resolv.conf /etc/resolv.conf` argument causes bwrap to fail. bwrap itself works fine in general on the affected machine but zerostack's invocation breaks.

**To Reproduce**
1. Use zerostack in WSL2 with bwrap as sandbox and have the model execute any bash command.
2. See error `bwrap: Can't create file at /etc/resolv.conf: No such file or directory`

**Expected behavior**
Sandboxed commands run successfully, with DNS resolution available (resolv.conf correctly exposed), regardless of whether /etc/resolv.conf is a regular file or a symlink to another filesystem.

**Operating System used**
Ubuntu inside WSL2 on Windows.

**LLM Provider and Model used**
any

**Suggestion**
It might be sufficient to use `std::fs::canonicalize` to get the real path of resolv.conf?
Alternatively, `--ro-bind-try` would skip the mount if it fails, then at least commands which do not DNS-resolve anything would work.

qznc:
In WSL /etc/resolv.conf can be a symlink to /mnt/wsl to get that information from the host. This broke zerostack using bubblewrap.

Fixes #115

### DIFF
--- a/src/sandbox.rs
+++ b/src/sandbox.rs
@@ -102,9 +102,23 @@ impl Sandbox {
             "/dev",
             "--tmpfs",
             "/tmp",
-            "--ro-bind",
-            "/etc/resolv.conf",
-            "/etc/resolv.conf",
+        ]);
+
+        match std::fs::canonicalize("/etc/resolv.conf") {
+            Ok(target) => {
+                cmd.arg("--ro-bind-try");
+                cmd.arg(target);
+                cmd.arg("/etc/resolv.conf");
+            }
+            Err(e) => {
+                tracing::warn!(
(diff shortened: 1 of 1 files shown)


gi-dellav:
Thanks for the contribution!

(P.s. you might want to join the matrix chatroom to get more development updates)

Cheers,
Giuseppe Della Vedova

PR #126, opened 2026-06-24
Nix code stable

### LINKED ISSUE #125: Nix code doesn’t work on stable; requires experimental flakes
**Describe the bug**
I don’t need to get into why flakes are a flawed design, but flakes _are_ an experimental feature. I wanted to try this project, but I don’t use & have disabled flakes since stable Nix works fine. Since there is no `flake.lock` there is no point in having something besides an overlay (which works on stable _or_ flakes-based Nix).

**To Reproduce**
1. Attempt to run the code with stable Nix

**Expected behavior**
Stable Nix can consume the project

**Screenshots**
n/a

**Operating System used**
NixOS

**LLM Provider and Model used**
None

**Additional context**
I will send a pull request that makes the Nix code modular. I have started & there are already issues where it doesn’t even build right as is.

toastal:
Flakes are experimental & shouldn’t be required for consumption. Convert to classic, stable Nix with an overlay being the *primary* way to access the code (both stable + flakes can use overlays). LLMs have been ‘poisoned’ with some bad Nix patterns—including not building stable Nix code first.

* added package-normal-form packages for the application + dev shell which make the code easier to consume add hoc as well as override
* fixed missing inputs for the packaging
* expose both thru overlays (with `development.nix` being only for development & not needed outside the project)
* use a release.nix to export the project
* left TODOs for questions that arose while refactoring
* dropped flake as it’s no longer providing any value—it isn’t even used to do input pinning without a `flake.lock`
* by proxy, eliminates `flake-utils` dependency on the Nix side… no dependencies, no experimental features

Used `nixpkgs-fmt` since this seems like this was the style of the existing code.

fixes #125


gi-dellav:
Thanks; honestly I don't know much about Nix, but the code seems correct except for one doubt that I have (but not using Nix); from what I understand just reading the code it seems like it's building from source by default, while the homebrew package downloads the binary by default, with the optional ability to build from source.

Can you please tell me how it works, and if I'm right, can you please set a binary-first default?

Cheers and thank you for contributing,
Giuseppe Della Vedova

p.s. you might want to join the Matrix chatroom to follow the development

toastal:
> I don't know much about Nix

I happen to teach it 😅

> from what I understand just reading the code it seems like it's building from source by default

Yes. This is what the Nix code here always done (you can tell be the `src = ../path/in/this/project`). This is generally how Nix prefers to do it since users can now do `zerostack.overrideAttrs (old: env.rustFlags = (old.env or { }) // { CARGO_BUILD_RUSTFLAGS = "-Ctarget-cpu=znver5"; }; })` for instance to change the build & its build flags. Pulling the built binary means Nix users need to trust your CI (we’ve seen a lot of Actions poisoning recently) but also you will need to be building it completely static if it isn’t (haven’t looked) since Nix/NixOS doesn’t follow FHS conventions requiring `patchELF` which at that point, it’d be better to build from source. The preferred caching mechanism in this case would be upstreaming a package into Nixpkgs & bumping Hydra (their CI) to build from source & add it to the giant binary cache. Even still, it’s often useful to have the Nix infracode around for user that want to test bleeding edge or do local development.

 > can you please set a binary-first default
 
This would normally be done with a `zerostack-bin` package (+ adding it to the default overlay) that could fetch that bin (I believe Firefox ships `*-bin` packages since it takes hours to build). If it’s statically built, fetching might be simple but generally this what Nix users would expect as a “default” unless this also takes hours to build (it didn’t).

Regardless, that would be better as a separate merge request since this one doesn’t change the existing functionality, just fixes minor errors, drops a dependency, extends to stable Nix support, & exposes an overlay for all users.

---

There’s still some room for improvements tho, but I didn’t want to dig into the project—especially: 1) proper source filtering for what the build minimally needs, 2) separating tests out between what can/can’t run in the sandbox (limited permissions, no network, & so on) so that `doCheck` can be true to make sure the tests run+pass 3) package definition more granular with multi-output targets.

gi-dellav:
Alright, I understand; I wanted to ask just one thing, could you please update the justfile (which is just a set of shell commands) to be compatible with the new Nix packaging?

Thanks,
Giuseppe Della Vedova

toastal:
Rebased… but what are you talking about?

```console
$ grep -i "nix" justfile || echo "NOT FOUND, bud."
NOT FOUND, bud.
```

toastal:
No `nix-command`s listed either?

```console
$ grep -e "nix (build|develop|shell)" (find . -type f)
```

toastal:
I guess I found a dangling comment that wasn’t doing anything.

gi-dellav:
You are right, I forgot that with the fact that I didn't really know how to manage Nix installations, so I left nix integration into our justfile to be done later.

Could you do it? You'd do me a great favour.

If not, I would still merge the PR in it's current stage as everything else seems correct.

toastal:
I’m not sure what the goal is… Checking if it builds? Delivering Nix is just delivering the ‘recipe’ to compile the code yourself.

gi-dellav:
It's just version switching (I don't know if it's needed on Nix, from what I searched only it wasn't needed for the Flakes version); for Homebrew we have to change the metadata to set the current version

gi-dellav:
Sorry, just saw that it's loaded from the TOML.

Cheers and thanks for the collaboration!

toastal:
https://codeberg.org/toastal/nixcfg/commit/e9b91768055928e2e64e8a9971075f2f5b2c7e8b

```console
$ zerostack --version
zerostack 1.5.0
```

👍‍

toastal:
```console
$ man zerostack
No manual entry for zerostack
```

bummer

gi-dellav:
Yeah, I should add manpage, for now you can run `less` on any file of `~/.local/share/zerostack/docs`

PR #127, opened 2026-06-24
feat(mcp): add OAuth 2.0 support for URL servers

wowi42:
## What

Adds OAuth 2.0 (authorization code + PKCE) support for URL-based MCP servers, using rmcp 1.7's `auth` module.

## Config

URL servers gain an optional `oauth` field. Use `true` for defaults (dynamic client registration, no extra scopes, loopback port 8970), or an object for explicit settings:

```json
{
  "mcp_servers": {
    "my-oauth-server": { "url": "https://example.com/mcp", "oauth": true },
    "scoped-server": {
      "url": "https://api.example.com/mcp",
      "oauth": { "scopes": ["read", "write"], "client_id": "abc", "redirect_port": 8970 }
    }
  }
}
```

## How it works

- Tokens persist per server under `<data_dir>/mcp-oauth/<server>.json` at mode 0600 (atomic write). Later sessions reconnect with the stored refresh token, no browser.
- The `CredentialStore` impl is hand-written (desugared async) so no `async-trait` crate is added; the loopback redirect uses `std::net::TcpListener` so no tokio `net` feature is needed.
- `/mcp login <server>` runs the flow in the TUI: prints the auth URL, catches the redirect, exchanges the code, then reconnects that server and rebuilds the agent live (no restart).
- `/mcp logout <server>` removes a stored token.
- OAuth URL servers connect through `AuthClient` + `StreamableHttpClientTransport::with_client`; non-OAuth paths are unchanged.

## Notes / tradeoffs

- Enables rmcp's `auth` feature (`rmcp?/auth`), which pulls in `oauth2` and a TLS/crypto backend (`aws-lc-rs`). Binary size grows somewhat; inherent to OAuth.
- During login the TUI event loop blocks until the redirect arrives (3 min timeout); not cancellable mid-wait. Could move to a background task later.
- Only authorization code + PKCE is implemented; client-credentials (M2M) is not wired up.

## Tests

12 new tests cover config parsing (`true` / object / disabled), redirect URI, filename sanitization, and callback parsing. Full suite: 475 pass with `mcp`, 463 without.

## Docs

`docs/CONFIG.md` documents the `oauth` field and login flow; `docs/COMMANDS.md` lists the new subcommands.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -661,7 +661,7 @@ dependencies = [
  "num-traits",
  "pastey 0.1.1",
  "rayon",
- "thiserror",
+ "thiserror 2.0.18",
  "v_frame",
  "y4m",
 ]
@@ -3304,7 +3304,7 @@ dependencies = [
  "reqwest 0.12.28",
  "serde",
  "serde_json",
- "thiserror",
+ "thiserror 2.0.18",
  "ureq",
  "windows-sys 0.60.2",
 ]
@@ -3919,7 +3919,7 @@ dependencies = [
  "jni-sys",
  "log",
  "simd_cesu8",
- "thiserror",
+ "thiserror 2.0.18",
  "walkdir",
  "windows-link",
 ]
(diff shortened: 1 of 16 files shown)


gi-dellav:
Seems all correct, I'll finish the review and merge it; it should come out with v1.5.1-rc1 later today.

Thanks for contributing!

p.s. you might want to join the Matrix chatroom to follow the development

PR #128, opened 2026-06-24
feat(ui): add show_cost_always to display session cost at zero

wowi42:
## What

Adds a `show_cost_always` config option (boolean, default `false`) that makes the status bar display the session cost even when it is `$0.0000`.

By default the cost is hidden until it rises above zero, so models with no per-token pricing configured never show a cost. With this option on, the cost is always visible.

```toml
show_cost_always = true
```

## How it works

- `Config.show_cost_always: Option<bool>` with `resolve_show_cost_always()` (default `false`).
- A non-persisted `#[serde(skip)]` flag on `Session`, set once from config at TUI startup, so the status renderer can apply it without threading config through every `refresh_display` call.
- `src/ui/status.rs`: cost is shown when `total_cost > 0.0 || show_cost_always`.

## Tests

5 new tests in `src/tests/status_tests.rs`: cost hidden by default at zero, shown when the flag is set, shown when nonzero regardless of the flag, and the resolver default + config parse. Full suite passes (468).

## Docs

`docs/CONFIG.md` documents the new key.


gi-dellav:
Thanks for contributing, just merged, will ship with v1.5.1-rc1 (later today)

P.s. you might want to join the Matrix chatroom to be updated on the development

PR #129, opened 2026-06-24
fix(session): refresh context window from catalog on resume

wowi42:
## What

Fixes a resumed session showing a stale context window. For example `deepseek/deepseek-v4-pro` (1M in the catalog) kept showing 128k after resume.

## Cause

`--continue` / `--session` load a saved session as-is, and `Session.context_window` is persisted. A session first saved before its model had a catalog entry (or saved on a different model) kept the old value forever, even after the catalog became correct. Fresh sessions, `/model`, and `/provider` all re-resolve correctly; only resume did not.

## Fix

- Add `Config::catalog_context_window(provider, model) -> Option<u64>`, which returns the catalog value only (no config override, no 128k fallback), so a real catalog hit is distinguishable from the default. `resolve_context_window` now builds on it.
- On session load (`main.rs`), if the user has not pinned `context_window` in config and the catalog knows the session's model, refresh the window. A pinned config value still wins everywhere; custom or unlisted models keep their persisted value (no accidental shrink); idempotent for fresh sessions.

## Tests

Catalog lookup for a known 1M model, `None` for unknown/no-catalog providers, and config-pin precedence over the catalog.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -222,17 +222,21 @@ impl Config {
         if let Some(cw) = self.context_window {
             return cw;
         }
-        if let Some(entries) = crate::models_catalog::catalog_entries(provider) {
-            for e in entries {
-                if e.id == model_id {
-                    if let Some(cl) = e.context_length {
-                        return cl as u64;
-                    }
-                    break;
-                }
-            }
-        }
-        128_000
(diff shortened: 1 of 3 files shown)


gi-dellav:
Thanks for contributing, just merged, will ship with v1.5.1-rc1 (later today)

P.s. you might want to join the Matrix chatroom to be updated on the development

PR #132, opened 2026-06-25
feat(ctx): accurate context figure and recover failed sends

xavierforge:
## Summary

Makes the status-bar context figure accurate before the first provider
response and stops a failed send from poisoning the session.

## Motivation

The main pain: sending an oversized message returns an error (e.g. `prompt is
too long`), but the message was already in history with no reply. It stayed
there and was rejected on every following turn, so the session was stuck until
a manual `/undo`. Investigating that surfaced two related context-figure
issues: the figure started at 0 and only became real after the first response,
and `/undo` left it stale.

## Changes

**Context estimate accuracy**
- Include the request preamble (system prompt, tool-use guidance, context
  files, memory) in the pre-calibration estimate. Extracted `build_preamble`
  from `build_agent_inner` and added `estimate_overhead`, seeded at startup and
  refreshed in `ensure_agent`. The figure now reflects the real baseline from
  the first frame instead of showing 0.
- A leading `~` marks the figure while it is still an estimate
  (`calibrated_tokens == 0`), dropped once the provider reports real usage, so
  the snap to the exact count reads as a refinement.

**Failed-send recovery**
- On any errored turn, the trailing pending user message is rolled back into
  the input editor instead of being left in history. This is provider-agnostic
  and needs no error classification: a failed turn produced no response, so
  rolling the message back to a pending state is the right recovery for
  too-long, auth, rate-limit, and transient errors alike. The user sees the
  error and decides whether to shorten and resend or fix the cause and resend
  unchanged.

**Context figure on removal**
- `Session::truncate_to`, shared by `/undo` and the rollback, subtracts a
  removed turn's estimated tokens from the calibration anchor instead of
  resetting it (which undercounts, since the estimate omits tool schemas) or
  leaving it stale (which overcounts). The figure tracks the real remaining
  size.

## Notes

- No new dependencies. `cargo build` clean.
- The per-message estimator is unchanged; EMA self-calibration was considered
  and left out as a display-only refinement that the `~` marker already covers.

### DIFF
--- a/src/agent/builder.rs
+++ b/src/agent/builder.rs
@@ -13,23 +13,12 @@ use crate::permission::ask::AskSender;
 use crate::permission::checker::PermCheck;
 use crate::sandbox::Sandbox;
 
-#[allow(clippy::too_many_arguments)]
-pub async fn build_agent_inner<M: CompletionModel + 'static>(
-    model: M,
-    cli: &Cli,
-    cfg: &Config,
-    context: &ContextFiles,
-    permission: Option<PermCheck>,
-    ask_tx: Option<AskSender>,
-    sandbox: Sandbox,
-    reasoning_enabled: bool,
-    temperature: Option<f64>,
-    // Provider-specific extra body params (e.g. OpenRouter `provider.order` to
-    // pin Claude to the Anthropic direct route so `cache_control` is honored).
-    // `None` for providers that need no extra routing.
-    additional_params: Option<serde_json::Value>,
-    #[cfg(feature = "mcp")] mcp_manager: Option<&McpClientManager>,
-) -> Agent<M> {
+/// Assemble the system-prompt preamble every request carries: the base
+/// `SYSTEM_PROMPT`, tool-use guidance, context files (AGENTS.md, ARCHITECTURE.md,
+/// active mode prompt), working directory, `/add`ed files, memory, and the user
+/// `SUFFIX.md`. Extracted from [`build_agent_inner`] so its token cost can be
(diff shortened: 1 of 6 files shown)


gi-dellav:
Thanks, will review and merge, do you know what is the cause of the "prompt is too long" issue? I recently sent a ~100 line prompt and there were no issues

xavierforge:
圖片

As demonstrated in the screenshot, sending a large prompt when already operating near the context window limit will trigger a 400 Bad Request error.

This specific error message, "prompt is too long: 208576 tokens > 200000 maximum", was observed when using Anthropic's API.
A similar issue occurs with OpenAI, which returns a context_length_exceeded error stating: "Your input exceeds the context window of this model. Please adjust your input and try again."

So the issue isn't just about the length of the single prompt you just sent, but the accumulated total of the entire conversation history.

I first encountered this kind of issue when using the Gemini CLI while adding a whole codebase in context.
And I trigger this while I was debugging compaction logic, while I do `cat src/ui/mod.rs src/main.rs src/provider.rs > /tmp/zs_big.txt` and paste the content of `zs_big.txt` in chat multiple times.

I first encountered this kind of issue when using the Gemini CLI while loading an entire codebase into the context.
And I triggered this again while debugging compaction logic a few days ago. 
I had concatenated several source files using `cat src/ui/mod.rs src/main.rs src/provider.rs > /tmp/zs_big.txt` and then pasted the content of `zs_big.txt` into the chat multiple times, which quickly exhausted the context window.

gi-dellav:
Alright, thanks, merged.

PR #134, opened 2026-06-26
fix(sandbox): must bind /etc/resolv.conf before /

qznc:


a-zw:
This actually fixes #115 for me.

gi-dellav:
Thanks, merged

PR #135, opened 2026-06-26
feat(ui): configurable left margin for the chat area

wowi42:
## What

Adds a `chat_left_margin` config option that left-pads the chat area by N columns. Only the chat buffer is affected - the input line and status bar stay flush against the edge.

```toml
chat_left_margin = 2
```

Default `0` (no change in behavior).

## How it works

- `src/ui/renderer.rs`: a `chat_margin` field (set via `set_chat_margin`, clamped to keep usable width). `max_line_width` and `buffer_line_at_row` subtract the margin so word wrapping and mouse-selection row mapping stay correct. `render_viewport` (the authoritative redraw) paints a margin-wide gutter in the chat background before each line; the live `write_line` and streaming `write` paths paint/offset by the margin too, so streaming text lands in the right column immediately.
- `src/config/mod.rs`: `chat_left_margin` key + `resolve_chat_left_margin()` (default 0), wired in `run_interactive` right after the renderer is created.

## Tests

`renderer_tests.rs` asserts the margin reduces content width and that 0 leaves it unchanged. Full suite passes.

## Note

CI may show the flaky `edit_tests::test_sim_preserves_crlf_line_endings` failure - that's the pre-existing edit-system test race fixed by #133, unrelated to this change.


gi-dellav:
Thanks for the fast contribution, merged!

PR #136, opened 2026-06-26
feat(input): scroll, click-to-position, and a height cap for tall input boxes

xavierforge:
## What

Makes a multi-line input box that is taller than the screen actually usable.
Previously, once the input grew past the visible area you could not see or reach
earlier content: the viewport stayed pinned to the bottom, Up/Down moved the
cursor off-screen, the mouse wheel only scrolled the chat (and got stuck), and
there was no way to click to a position.

## Changes

- **Cursor-following viewport.** The input now scrolls vertically to keep the
  cursor line visible, so Up/Down reveal lines that do not fit at once instead
  of the view staying anchored to the bottom.
- **Mouse-wheel scrolling.** The wheel scrolls the input viewport first, then
  continues back through the chat history once the input is at its top;
  scrolling down brings the chat back before the input.
- **Click to position the cursor.** Clicking inside the input box moves the
  cursor to that spot (wide chars, horizontal and vertical scroll accounted
  for). Clicks outside still start a chat-history text selection.
- **Height cap (~30%).** The input never uses more than roughly 30% of the
  available area, so the chat history stays visible and scrollable above a tall
  input instead of being squeezed to nothing.
- **Caret clamp.** The terminal caret is clamped to the visible input rows so it
  never spills onto the separator or status bar while scrolled away.
- **Redraw sync.** The input height is reconciled before the chat viewport is
  drawn, so changing the input height no longer leaves a stale separator (when
  it shrinks) or chat text hidden under the input (when it grows) until the next
  redraw.

## Files

- `src/ui/renderer.rs` — viewport/height/caret/geometry logic and the
  click-to-cursor mapping.
- `src/ui/mod.rs` — wheel and mouse-down event wiring; height sync before
  rendering the chat.
- `src/ui/input/mod.rs` — `set_cursor` for click placement.

## Testing

- `cargo build` and `cargo test` pass.
- Manually: typing/pasting a long multi-line message, then Up/Down, mouse wheel
  (into chat history and back), clicking to reposition, and deleting all text to
  confirm the separator follows immediately.

### DIFF
--- a/src/ui/input/mod.rs
+++ b/src/ui/input/mod.rs
@@ -58,6 +58,18 @@ impl InputEditor {
         }
     }
 
+    /// Move the cursor to `pos` (a byte offset), clamped to a char boundary
+    /// within the buffer. Used when a mouse click places the cursor.
+    pub fn set_cursor(&mut self, pos: usize) {
+        let pos = pos.min(self.buffer.len());
+        self.cursor = if self.buffer.is_char_boundary(pos) {
+            pos
+        } else {
+            prev_char_boundary(&self.buffer, pos)
+        };
+        self.yank_pos = None;
+    }
+
     pub fn clear_buffer(&mut self) {
         self.buffer.clear();
         self.cursor = 0;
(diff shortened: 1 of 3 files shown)


gi-dellav:
Nice, thank you, will review and merge later today

PR #141, opened 2026-06-27
Forward reasoning deltas from streamed responses

fayalalebrun:
## Summary

Forwards reasoning deltas from streamed model responses so reasoning output reaches the UI/event pipeline.

## Context / Motivation

Some providers stream reasoning separately from normal assistant text. Dropping those deltas makes reasoning-capable responses appear incomplete or inconsistent.

## Key Changes

- Handles reasoning stream deltas in the agent runner.
- Emits reasoning events through the existing event pipeline.

## Verification & Testing

Reviewers can use a streaming reasoning-capable model and confirm reasoning text appears as it is streamed.


gi-dellav:
Could you please add a `show_reasoning` config field so that behaviour is configurable?

Thanks in advance

fayalalebrun:
Correction: added `show_reasoning` config support and force-pushed the branch. It defaults to the existing visible behavior (`true`) and can be disabled with `show_reasoning = false`.

PR #143, opened 2026-06-27
Persist tool outputs in session history

fayalalebrun:
## Summary

Persists tool outputs in saved session history so restored conversations retain the tool results that informed prior assistant turns.

## Context / Motivation

When tool results are missing after restore, the transcript is less useful and future turns may lack important context about what happened earlier.

## Key Changes

- Records tool output data in session state.
- Restores persisted tool outputs when loading sessions.
- Adds session storage coverage for tool output persistence.

## Verification & Testing

Reviewers can run a tool call, save and restore the session, and confirm the restored transcript still includes the tool result.


gi-dellav:
Could you please rebase this on main after I merge your other open PR?

Thanks in advance

PR #147, opened 2026-06-28
Add direct DeepSeek provider support

fayalalebrun:
## Summary

Adds DeepSeek as a first-class provider with model catalog entries and direct API-key resolution.

## Context / Motivation

DeepSeek can be configured as an OpenAI-compatible custom provider, but a first-class provider removes setup friction and makes the default model/catalog experience clearer for users who want direct DeepSeek access.

## Key Changes

- Adds DeepSeek provider selection and client construction.
- Adds `DEEPSEEK_API_KEY` resolution through the existing provider auth path.
- Adds DeepSeek model catalog entries.
- Adds provider/auth tests for DeepSeek behavior.

## Verification & Testing

Reviewers can set a DeepSeek API key, select the DeepSeek provider, start a session with a DeepSeek catalog model, and confirm missing-key behavior matches other direct providers.

### DIFF
--- a/data/models.json
+++ b/data/models.json
@@ -81,6 +81,18 @@
       "context": 1000000
     }
   ],
+  "deepseek": [
+    {
+      "id": "deepseek-v4-flash",
+      "name": "DeepSeek V4 Flash",
+      "context": 1000000
+    },
+    {
+      "id": "deepseek-v4-pro",
+      "name": "DeepSeek V4 Pro",
+      "context": 1000000
+    }
+  ],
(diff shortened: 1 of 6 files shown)


gi-dellav:
Hi, sorry for the delay: bring minimal, we don't really want to implement directly in the sou ever code new providers that are compatible with the OpenAI API, but if you are interested there were in the past discussions about adding a simpler system for managing commonly used custom providers (like a backed-in list of providers); we can talk more about that on Element

PR #148, opened 2026-06-28
Store long tool outputs as sidecar artifacts

fayalalebrun:
## Summary

Moves large tool outputs into sidecar artifact files while keeping references in the session.

## Context / Motivation

Large command/tool outputs can bloat session JSON and context. Sidecars preserve access to the full output without forcing the entire payload into the main session file.

## Key Changes

- Detects long tool outputs.
- Writes oversized outputs to sidecar files under session storage.
- Stores compact references in session history.
- Keeps shorter outputs inline for readability.

## Verification & Testing

Reviewers can run a command that produces a large output, save the session, and confirm the session references a sidecar file containing the full output.

### DIFF
--- a/src/session/mod.rs
+++ b/src/session/mod.rs
@@ -1,9 +1,15 @@
 pub mod chat_history;
 pub mod storage;
 
+use std::path::Path;
+
 use compact_str::CompactString;
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
+
+pub const TOOL_RESULT_SAVE_THRESHOLD: usize = 12_000;
+pub const TOOL_RESULT_HEAD_CHARS: usize = 2_000;
+pub const TOOL_RESULT_TAIL_CHARS: usize = 8_000;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "snake_case")]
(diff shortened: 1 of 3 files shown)


gi-dellav:
Makes sense, do you think we should implement a garbage collector logic for sessions and/or sidecars?

fayalalebrun:
Based on my preference I would not do it for sessions as I sometimes find it useful to reference old sessions. Claude Code does delete sessions older than 30 days, but OpenCode does not for example.

So far I have been using this for about a week pretty heavily and I've only accumulated around 250 MB. Way less than most compiler artifacts accumulated in the same period. If it becomes an issue we could  implement a utility for the users to call when desired. Like: `zerostack garbage-collect --older-than=30d`.

gi-dellav:
Just out of curiosity, can you tell me how many session files were generated in that period?

It's just to get an estimate, if the size-per-session is low enough I won't add garbage collection and will merge this PR in its current state.

fayalalebrun:
I have 89 session files generated during the period. Though just to be clear this change should only decrease the size of the sessions themselves (since large tool outputs are moved to sidecar files). There will be a marginal increase overall because of the repetition between the two, around 8K characters per tool output above 12K characters. The 250 MB is counting both session files and sidecar tool outputs.

gi-dellav:
Makes sense, I will merge it without garbage collection for now, maybe I'll add it in a future release.

Thanks for contributing

PR #149, opened 2026-06-28
Add conversation forking

fayalalebrun:
## Summary

Adds conversation forking so users can branch a new session from an earlier point in an existing conversation.

## Context / Motivation

Forking is useful when a user wants to explore an alternate direction without losing or mutating the original session.

## Key Changes

- Adds session helpers for deriving fork titles and truncating messages at a fork point.
- Adds `/fork` handling and a TUI picker over prior user turns.
- Resets cost/token/calibration state for the new fork while preserving the selected prior context.

## Verification & Testing

Reviewers can create a multi-turn session, use `/fork`, select a prior user turn, and confirm the new session contains only the expected prior messages while the original session is unchanged.

### DIFF
--- a/src/session/mod.rs
+++ b/src/session/mod.rs
@@ -281,6 +281,56 @@ impl Session {
         self.updated_at = CompactString::new(chrono::Utc::now().to_rfc3339());
     }
 
+    pub fn title(&self) -> String {
+        if !self.name.is_empty() {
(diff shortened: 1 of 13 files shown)


fayalalebrun:
Closing mostly superseded by https://github.com/gi-dellav/zerostack/pull/151

PR #150, opened 2026-06-28
Flush atomic temp files before rename

fayalalebrun:
## Summary

Flushes atomic-write temp files before renaming them into place.

## Context / Motivation

CI has been intermittently observing torn reads in the atomic write test. The temp file was written and then renamed after the async file handle scope ended, but the buffered writes were not explicitly flushed first.

## Key Changes

- Calls `flush()` after writing the temp file and before `rename`.

## Verification & Testing

- `cargo test no_torn_reads_during_rewrites -- --nocapture`
- `cargo test test_hash_range_edit -- --nocapture`


gi-dellav:
Thanks

PR #151, opened 2026-06-28
feat: /rewind to an earlier turn and a reversible /undo

xavierforge:
## Summary

Adds a way to rewind the conversation and makes `/undo` reversible:

- **`/rewind`** opens a picker over earlier user turns. Pick one, confirm "Rewind and start from here", and the conversation rewinds in place to just before that turn, with the turn loaded back into the input box for editing.
- **`/redo`** restores the messages removed by the last `/undo` or rewind.

## Behavior

- `/undo` keeps working as before, but is now reversible with `/redo`.
- `/rewind` opens a two-level picker: pick a turn, then confirm "Rewind and start from here" or "Cancel". Esc within the picker backs out, consistent with the other pickers. Nothing changes until you confirm.

## Design notes

- **One rewind primitive.** `/undo` and `/rewind` both go through `Session::rewind_to`, which records a single-step restore point and then truncates. Rather than add a parallel "undo of a rewind" mechanism, this generalizes the existing one, so `/undo` gains `/redo` for free. It reuses the existing calibration-aware `truncate_to`, so the context figure stays accurate after a rewind instead of going stale.
- **Restore point is single-step.** It is dropped as soon as the conversation moves forward again (a new message), so `/redo` only ever reverses the cut it directly follows.
- **Rewind is conversation-only.** It never touches the git working tree; the conversation and the code state stay independent.
- **The chosen turn is loaded back into the input** for editing, since the common case is revising that message rather than starting from a blank line.
- **No new key bindings.** Exposed as a slash command so it does not reserve Esc (or any other key), which a host TUI may use for its own navigation.
- **Useful as a context-management tool.** Jumping back to a clean point after an off-topic tangent reclaims the context that the tangent consumed (the truncation is calibration-aware, so the figure drops correctly), then you re-add just the insight you want to keep. It complements `/compress`: a surgical cut rather than a summary.

## Testing

- Unit tests for the rewind primitive (`rewind_to`, `redo`, restore-point invalidation) and the two-level picker state machine (confirm, cancel, back-out).
- Full suite green; clippy clean.
- Verified live in the TUI: `/rewind`, rewind to an earlier turn, edit the loaded message, and `/redo` to restore, plus the picker back-out paths.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8759,7 +8759,7 @@ checksum = "e13c156562582aa81c60cb29407084cdb54c4164760106ab78e6c5b0858cf64e"
 
 [[package]]
 name = "zerostack"
-version = "1.5.1-rc1"
+version = "1.5.1-rc2"
 dependencies = [
  "agent-client-protocol",
  "anyhow",
--- a/src/session/mod.rs
+++ b/src/session/mod.rs
@@ -29,6 +29,19 @@ pub struct SessionMessage {
     pub estimated_tokens: u64,
 }
 
+/// A single-step restore point captured before a conversation rewind, so the
+/// destructive truncation can be undone with `/unrewind`. Holds the full
+/// message list and the calibration/estimate fields that `truncate_to` mutates,
+/// which is everything needed to put the session back exactly as it was. Not
+/// persisted: a rewind is only undoable within the session that made it.
+#[derive(Debug, Clone)]
+pub struct RewindUndo {
+    messages: Vec<SessionMessage>,
+    total_estimated_tokens: u64,
+    calibrated_tokens: u64,
+    calibrated_msg_count: usize,
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Compaction {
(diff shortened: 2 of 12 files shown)


gi-dellav:
I noticed a TUI issue: using `Esc` to implement any behaviour in zerostack brings to a conflict with `multistack` which uses `Esc` as the core key for going back to the agents dashboard.

xavierforge:
Good catch on the Esc conflict with multistack. I reworked it to avoid Esc entirely:

- Dropped the Esc bindings (single Esc interrupt and double-Esc rewind). `Ctrl+C` still interrupts a running turn, so nothing is lost there.
- The rewind is now a `/rewind` slash command that opens the same two-level picker. zerostack no longer binds Esc (or any new key) for this, so it won't collide with the dashboard navigation.

The `/undo` -> `/redo` part is unchanged, since it never touched Esc. PR and description are updated, and I re-verified the `/rewind` flow live (rewind to an earlier turn, edit the loaded message, `/redo` to restore).

PR #152, opened 2026-06-28
Add YAML config parser

m-u-xyz:
YAML is a superset of JSON and improves editability for users (no braces, no trailing commas).

Since serde_yaml_ng is in zerostack's transitive dependencies anyway, supporting it does not increase its footprint.


gi-dellav:
What is the advantage of using YAML over both JSON and TOML?

m-u-xyz:
yaml's indents show the actual data hierarchy.

Frankly, my brain more-or-less refuses to work with TOML. (Unfortunately this is not as exaggerated as I'd like it to be, given the reliance on toml by both Python and Rust infrastructure.)

Also, yaml requires less quoting of literals than both others.

gi-dellav:
Makes sense, the edit is small enough, thanks for the contribution

PR #153, opened 2026-06-28
Retrying and ACP didn't talk to each other.

m-u-xyz:
The AgentEvent enum gained a Retrying variant, but the match event loop in the ACP server didn't handle it.


gi-dellav:
thanks, merged

PR #155, opened 2026-06-28
ci: pin CI/CD versions, harden releases (rc prereleases), enforce lint

wowi42:
## What

Harden CI/CD: pin every version, fix release handling for rc tags, and enforce the lint bar in CI.

## Version pinning

- Every GitHub Action pinned to a full commit SHA with a `# vX.Y.Z` comment, at the latest stable major: checkout v7.0.0, setup-rust-toolchain v1.17.0, taiki-e/install-action v2.82.5, configure-pages v6.0.0, upload-pages-artifact v5.0.0, deploy-pages v5.0.0, create-pull-request v8.1.1.
- New `rust-toolchain.toml` pins Rust 1.96.0 with rustfmt and clippy; workflows also set `toolchain: "1.96.0"` explicitly.
- New `.github/dependabot.yml` for github-actions and cargo (weekly, grouped, conventional-commit prefixes) so the pins stay current.

## Releases

- rc tags (any hyphen suffix, e.g. `v1.5.1-rc2`) now publish as GitHub prereleases instead of latest.
- A single `create-release` job creates the release once via `gh` with generated notes, removing the race where both matrix jobs created it.
- Assets upload via `gh release upload` instead of a third-party action; a `SHA256SUMS` file is published; `fail-fast` is disabled so one target failing still ships the rest.
- `permissions` narrowed from `write-all` to `contents: write`.

## CI

- New `lint` job: `cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings` (same bar as the justfile and pre-commit hook, previously not enforced in CI).
- Concurrency added; cancels in-progress PR runs.
- Pre-existing clippy violations in test modules fixed so the new gate is green (pure refactors, no behavior change).

## Incidental

- `Cargo.lock` synced from 1.5.1-rc1 to 1.5.1-rc2 to match the manifest.

## Verification

- `cargo fmt --check`: pass
- `cargo clippy --all-targets --all-features -- -D warnings`: pass
- `cargo test`: 533 passed

## Not included

- The `justfile` `add-tag` recipe pushes `origin dev`, but every workflow triggers on `main`. Left untouched in case it is intentional. Can reconcile in a follow-up.
- The action major-version bumps (checkout, the Pages trio, create-pull-request) have not yet run in CI. The inputs used are unchanged across these majors.


gi-dellav:
Thanks for the contribution!

Issue #137, opened 2026-06-26
Allows sandbox access to ~/.cache


qznc:
For me specifically, bazel is the challenge. It builds everything in a sandbox which is located  in `~/.cache/bazel-cache/`. With bubblewrap it just fails. This is needed for rather basic things like running tests, so I want the agent to do it itself.

However, for sure bazel is not the only tool which needs something writable somewhere such that `/tmp` does not cut it.

**Describe the solution you'd like**

The simplest solution would be to always bind `~/.cache/` as writeable with bwrap.

**Describe alternatives you've considered**

For more fine-grained and flexible control, it might be a better idea to allow configuration arbitrary writable folders. Maybe even per project?

Alternatively, maybe zerostack is not really the right place and it should actually be bubblewrap we want to configure directly somehow?

gi-dellav:
Simple enough, will fix later today, thanks for the feedback

toastal:
Does it do `$TMPDIR`? OpenCode drives me nuts that it’s always asking for full access to `/tmp` which contains sensitive stuff instead of using the proper env var.

gi-dellav:
@toastal  No it doesn't, but it's a good idea; after dinner I'll update the sandbox system both to allow ~/.cache by default and if $TMPDIR is found, use that instead of the hardcoded `/tmp`

toastal:
There’s also the `$XDG_RUNTIME_DIR` which is usually tmpfs, which could be a better scratchpad since it doesn’t persist on reboot, & doesn’t write to disk to wear the drive. (This is where I set my per-project `$TMPDIR` using direnv, but the harnesses never check `$TMPDIR`, but this assumes you have spare RAM).

qznc:
zerostack [currently does](https://github.com/gi-dellav/zerostack/blob/df21b8d1fd21f44c1ea1d693af650ad643b4f9e2/src/sandbox.rs#L117) `--tmpfs /tmp` with bubblewrap, but I'm not quite sure what that means. Is it a fresh tmpfs per invocation?

The `~/.cache` is meant for longer-term things which should persist across reboots, so tmpfs would not be the right approach for that.

gi-dellav:
Yes, `--tmpfs` mounts a new clean tmpfs in a path recognized by the sandbox as `/tmp`.

[...]

PR #159, opened 2026-06-30
fix(sandbox): must bind CACHE HOME after /

qznc:
Another fix related to #137 

To be honest, I don't really understand bwrap here. The order is relevant but apparently not documented anywhere.


gi-dellav:
Thanks, hope that sandbox will finally work; I honestly don't use it, but if there are other issues, contact me on Element and I'll recheck the bubblewrap documentation

PR #165, opened 2026-07-06
fix: security and robustness fixes from code review

wowi42:
Critical:
- Bash permission: newline is now a command separator, preventing bypass where a denied command was hidden after an allowed one
- Permission: deny rules evaluated before session allowlist and allow_all_mcp_calls, so neither can bypass an explicit deny

High:
- Atomic session/history writes (temp+rename) to prevent corruption on crash; corrupt chat history is backed up instead of erased
- save_quick_model returns error on parse failure instead of wiping config via unwrap_or_default()
- is_external_path uses canonicalize() to resolve symlink traversal
- Invalid regex in deny rules falls back to match-all (fail-safe) and logs a warning instead of silently degrading
- Markdown re-parse throttled to line boundaries on streaming tokens (O(n^2) -> O(n))
- 64KB cap on ancestor context files (AGENTS.md/CLAUDE.md)
- clear_shrunk_rows uses saturating_sub to avoid u16 underflow
- /add files capped at 256KB with char-boundary-safe truncation

Medium:
- Doom-loop detection uses 8-call sliding window to catch alternating A,B,A,B patterns, not just consecutive exact repeats
- Stale prebuild agent no longer overwrites a live agent (is_none guard)
- ACP: auto-approves Ask requests instead of dropping the receiver; errors surface as Refusal stop reason + message
- MCP: shutdown() calls running_service.cancel() instead of drop
- Runner caps empty-response loop at 3 iterations
- git add -u (tracked only) instead of -A to avoid committing secrets
- Pure-Rust PATH search for sandbox binary (no which dependency) + startup warning when sandbox unavailable
- ChdirGuard RAII for panic-safe CWD restore in git_worktree
- enable_exa_mcp defaults to false

Low:
- Char-boundary-safe string truncation (runner, session)
- unreachable!() in MIME type matching replaced with default+warn
- Ponytail comment documenting deferred async git migration

All 579 tests pass, no warnings.

### DIFF
--- a/src/agent/builder.rs
+++ b/src/agent/builder.rs
@@ -65,14 +65,31 @@ pub fn build_preamble(context: &ContextFiles, reasoning_enabled: bool) -> String
 
     let total_len = total_len + suffix.as_ref().map_or(0, |s| s.len() + 6); // "\n\n---\n\n"
 
-    // Add extra files content to preamble budget
+    // Add extra files content to preamble budget. Cap each file to prevent a
+    // huge file from blowing up the system prompt past the context window.
+    const MAX_EXTRA_FILE_BYTES: usize = 256_000;
     let extra_files_content: Vec<String> = context
         .extra_files
         .iter()
         .filter_map(|p| {
-            std::fs::read_to_string(p)
-                .ok()
-                .map(|content| format!("Content of {}:\n{}", p.display(), content))
+            let content = std::fs::read_to_string(p).ok()?;
+            let truncated = if content.len() > MAX_EXTRA_FILE_BYTES {
+                tracing::warn!(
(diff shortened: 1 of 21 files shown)


gi-dellav:
Everything look correct, except for some edits that change the design of the agent:
- Why does it change the doom loop system? The only issue is with consecutive calls, ABAB patterns are accepted and common (ex. edit->bash->edit->bash)
- Why does it set Exa MCP to false? The README advertises the Exa as default, exactly like Opencode and Mistral's Vibe, in order to offer `websearch` and `webfetch`

Everything else looks correct and is ready to be merged; thank you again for this big PR.

PR #166, opened 2026-07-07
feat: named sessions — --name flag, /rename command, resume by name

wowi42:
## Summary

Adds named sessions to zerostack.

## Changes

- **`--name <name>`** CLI flag — set a session name at creation
- **`/rename <name>`** slash command — rename the current session in-TUI
- **`--session`** now matches by name (prefix/substring) when ID is ambiguous, plus a separate exact-name recovery path
- Session listings (CLI `--resume` and TUI `/sessions`) show `[name]` for named sessions
- Updated docs: COMMANDS.md, GET_STARTED.md, index.html
- Added `/rename` to the slash command picker

## Usage

```bash
zerostack --name "my-bugfix"          # create named session
zerostack --session my-bugfix          # resume by name
```

```
/rename refactor-auth                  # rename in-TUI
/sessions                              # now shows [name]
```

### DIFF
--- a/docs/COMMANDS.md
+++ b/docs/COMMANDS.md
@@ -11,8 +11,9 @@ All slash commands are available from the TUI input prompt.
 | `/retry` | Load the last user message into the input editor for editing. |
 | `/quit` | Exit zerostack. |
 | `/sessions` | List recent saved sessions (up to 20). |
-| `/sessions <id-prefix>` | Load a session by its ID prefix. |
-| `/sessions delete <id-prefix>` | Delete a session by its ID prefix. |
+| `/sessions <id-or-name>` | Load a session by its ID prefix or name. |
+| `/sessions delete <id-or-name>` | Delete a session by its ID prefix or name. |
+| `/rename <name>` | Rename the current session. |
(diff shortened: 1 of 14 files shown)


gi-dellav:
Makes sense, later today I'll review and merge

gi-dellav:
All good, except for one question: what is the name/ID given to unnamed sessions?

I couldn't understand it from the diffs of this PR.

Thanks,
Giuseppe Della Vedova

wowi42:
Hi Giuseppe, good catch - the PR description didn't make this explicit.

For unnamed sessions (when the user does not pass `--name`):

- **ID:** a UUID v4, generated in `Session::new` at `src/session/mod.rs:169`. This is the unique key used for the on-disk file (`<id>.json`) and for session matching.
- **Name:** an empty string ``, set in `src/main.rs:166` via `cli.name.as_deref().unwrap_or()`.

So the session still has a stable UUID identity, but `session.name` stays empty and the UI treats it as unnamed: no `[name]` label in `/sessions` listings, and `/rename` reports "current session has no name".

If you'd prefer unnamed sessions to get a more visible default label, happy to add that in a follow-up.

PR #175, opened 2026-07-08
fix: mouse text selection broken by missing bottom-alignment padding

wowi42:
## Problem
Click-and-drag text selection in the chat viewport was broken — clicks on the content area returned \`None\` from \`buffer_line_at_row()\`, so no selection was ever started.

## Root Cause
\`buffer_line_at_row()\` started \`visual_row\` at 0, but the viewport renders padding rows at the top before actual content when auto-scrolled and content is shorter than the viewport (bottom-alignment). This meant clicks in the content area were off by \`pad\` rows.

## Fix
Start \`visual_row\` at \`pad\` (instead of 0) to match the rendering offset. Also hoisted the \`pad\` computation so it's available for both the early-return guard and the \`visual_row\` initial value.

### DIFF
--- a/src/ui/renderer.rs
+++ b/src/ui/renderer.rs
@@ -232,11 +232,13 @@ impl Renderer {
         }
 
         let auto_scroll = self.scroll_offset == 0;
-        if auto_scroll && total < visible {
-            let pad = visible - total;
-            if (row as usize) < pad {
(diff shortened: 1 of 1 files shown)


gi-dellav:
Thank you, merged!

PR #178, opened 2026-07-09
feat(hooks): external-command hooks at agent lifecycle points

xavierforge:
## Summary

Adds an optional `hooks` feature (default-off) that lets external commands observe or gate the agent at defined lifecycle points: a tool call, a user prompt, the agent finishing a turn, a session starting or ending, and a subagent starting or stopping. It follows the hook contract that Claude Code and Codex share, the same `settings.json` shape, stdin envelope, and exit-code / stdout-JSON convention, so hook scripts written to that convention mostly drop in with no changes.

## Motivation

Two things users could not do before: (1) gate or shape tool calls from outside the binary, deny a dangerous command, force a confirmation, rewrite arguments, or inject context, all without patching zerostack; and (2) react to the agent from other tools, notify an external system when a turn starts or finishes, keep a heartbeat, or push activity to a device.

Rather than invent a new format, this speaks the established hook contract, so scripts written for that convention work against zerostack unchanged. Verified end to end against Claude Code's contract: an unmodified Claude Code hook client drives a physical desk device off a terminal zerostack session.

## Trigger points

Nine events, chosen to cover every boundary the agent actually crosses:

| Boundary | Events |
| --- | --- |
| Tool call | `PreToolUse`, `PostToolUse`, `PostToolUseFailure` |
| Turn | `UserPromptSubmit`, `Stop` |
| Session | `SessionStart`, `SessionEnd` |
| Subagent | `SubagentStart`, `SubagentStop` |

`PostToolUse` and `PostToolUseFailure` are split on purpose: a tool that runs but returns a non-zero command exit is a success at the tool layer (`PostToolUse`), while a tool that itself errors is `PostToolUseFailure`. `PreCompact` and `Notification` are not implemented yet (natural follow-ups).

## What a hook can do

- `PreToolUse` is the only permission-blockable event. A handler's stdout JSON may set `permissionDecision` to `deny`, `ask`, or `allow`, or defer. `deny` holds even under `--yolo`; `ask` forces a confirmation and escalates to deny in non-interactive contexts (`-p`, `--loop`); `allow` only suppresses the prompt for that one call and can never override a deny from a rule, security mode, managed policy, or another hook. `PreToolUse` may also rewrite the tool arguments via `updatedInput`.
- `PostToolUse` may rewrite the model-visible result.
- `UserPromptSubmit` and `SubagentStart` may prepend `additionalContext`.
- `Stop` and `SubagentStop` may return `decision: block` with a `reason` to make the agent keep going instead of finishing (`Stop` gives up after 8 consecutive blocks without progress).
- Any handler may signal by exit code instead of JSON: `0` no objection, `2` blocks with stderr as the reason, anything else is a non-blocking error.

## Design decisions

- Shared hook contract on purpose, but its own file. Same schema, envelope, and contract, but zerostack reads its own `settings.json`, not any single tool's config file, so it does not couple to another product's config layout. Copying or symlinking the `hooks` key across works unchanged.
- rig-free dispatcher. The subsystem does not depend on rig's hook mechanism. Tools are wrapped by a decorator (`HookedTool`) and the lifecycle seams call the dispatcher directly, which keeps it isolated and unit-testable without a live agent or API key.
- Zero cost when off. Everything is behind the default-off `hooks` feature; tool wrapping is skipped entirely when no hooks are configured.
- Trust model. Global and managed hooks are trusted; project hooks (`.zerostack/settings.json`) require a one-time confirmation keyed by a hash of the handler definition, and are skipped with a warning in headless mode.
- Merged, deterministic verdicts. Multiple handlers on one event fold in declared order; the merge is what `--hooks-test` prints.

## User story

I keep a `PreToolUse` hook that denies any `Bash` command matching a destructive pattern, so it is blocked even when I run in `--yolo`. I add a `UserPromptSubmit` hook that injects the current git branch into context. Then I copy the hook block from my existing agent `settings.json` straight into `.zerostack/settings.json`, confirm it once, and my terminal agent now drives the same external tooling my other agent sessions did.

## Testing

- New hooks test suite covering the dispatcher, envelope building, tool-name normalization, subprocess timeout / process-group kill, the trust store, the permission and stop gates, and session / subagent lifecycle dispatch.
- A contract corpus: `envelope.json` + `hook.sh` + `expected.json` fixtures pinning the observable behavior (deny / ask / allow, `updatedInput` rewrite, stop block, exit-2 blocks, mixed-channel ignores JSON, no-decision defers).
- `cargo test --all-features` green (747 tests).
- `--hooks-test <tool>` dry-runs `PreToolUse` and prints the merged verdict with no session or API key, for manual verification.

## Reviewing

Split into five focused commits: the dispatcher subsystem, then its tests and corpus, then the runtime wiring, then the `/hooks` slash command, then the docs (`CONFIG.md`, `COMMANDS.md`).

## Notes

- Default-off feature; a prebuilt binary needs `--features hooks` (or `--all-features`) for any of this to exist.
- Pre-existing `collapsible_if` clippy findings in unrelated files are not touched here; they predate this branch and are outside its scope.


src/extras/hooks/trust.rs
```diff
@@ -0,0 +1,220 @@
...
+#[cfg(target_os = "linux")]
+fn managed_settings_path() -> PathBuf {
+    PathBuf::from("/etc/zerostack/managed-settings.json")
+}
+
+#[cfg(target_os = "macos")]
+fn managed_settings_path() -> PathBuf {
+    PathBuf::from("/Library/Application Support/zerostack/managed-settings.json")
+}
+
```
gi-dellav, line 70: Need windows support (even if experimental), following docs/CONFIG.md

src/provider.rs
```diff
@@ -629,23 +704,73 @@ impl AnyAgent {
         }
     }
 
-    pub fn spawn_runner(
+    pub async fn spawn_runner(
```
gi-dellav, line 705: Explain why it went from synchronous to async

src/extras/hooks/decorator.rs
```diff
@@ -0,0 +1,168 @@
...
+impl ToolDyn for HookedTool {
+    fn name(&self) -> String {
+        self.inner.name()
+    }
+
+    fn definition<'a>(&'a self, prompt: String) -> WasmBoxedFuture<'a, ToolDefinition> {
+        self.inner.definition(prompt)
+    }
+
+    fn call<'a>(&'a self, args: String) -> WasmBoxedFuture<'a, Result<String, ToolError>> {
```
gi-dellav, line 60: Explain the choice and usage of WasmBoxedFuture

src/agent/runner.rs
```diff
@@ -526,12 +566,26 @@ where
...
+/// `Stop` hook needs to do. The agent's own `default_max_turns` (set at
+/// construction, see `agent::builder::build_agent_inner`) still bounds
+/// internal tool-call round trips per call, same as [`spawn_agent`], which
+/// never used `.multi_turn()` either.
+#[allow(clippy::too_many_arguments)]
 pub async fn run_print<M, P>(
     agent: &Agent<M, P>,
     prompt: &str,
-    max_turns: usize,
+    _max_turns: usize,
```
gi-dellav: What happened max_turns?

src/agent/runner.rs
```diff
@@ -311,6 +311,11 @@ pub fn spawn_agent<M, P>(
     prompt: String,
     history: Vec<Message>,
     retry_config: RetryConfig,
+    // `--loop` iteration/active state, for the `Stop` hook envelope's
+    // `loop_iteration`/`loop_active` fields (per-iteration reset of
+    // `stop_hook_active`/the block cap falls out for free: each iteration is
+    // a fresh call to this function). `None` outside loop mode.
+    #[cfg(feature = "hooks")] loop_info: Option<(u32, bool)>,
```
gi-dellav: I don't really understand what loop_info contains, try to move it to a documented struct

gi-dellav (changes requested):
Just a few clarifications and small edits

src/extras/hooks/trust.rs
xavierforge: Fixed in 4858d5a. Added a Windows managed-settings path (`C:\ProgramData\zerostack\managed-settings.json`, mirroring the ProgramData convention for machine-wide policy) and kept the existing fallback for other Unixes. While auditing this I found one more Windows gap: the shell-form hook invocation hardcoded `sh -c`; it now follows the powershell -Command convention already used elsewhere in the codebase. The global path already resolves correctly on Windows via `dirs::config_dir()` (`%APPDATA%\zerostack\settings.json`), and the process-group handling in sandbox.rs was already unix-gated. CONFIG.md now lists the Windows paths, marked experimental.

src/provider.rs
xavierforge: `UserPromptSubmit` hooks can block the prompt entirely or rewrite it, so the dispatch result decides whether the runner is spawned at all (`Blocked` returns a feedback-only runner without touching the model) and with what prompt. That gate has to resolve before spawning, and dispatch is inherently async (tokio subprocess plus timeout). The alternatives were `block_on` inside a sync fn (panics inside the tokio runtime) or moving the gate into the spawned task (loses the never-spawn-on-block semantics). All call sites were already in async contexts, so the change is mechanical `.awaits`; `spawn_runner` was chosen as the seam precisely because all 9 UI/ACP submission paths already funnel through it, keeping the dispatch logic in one place. Added a doc comment in a24bb59.

src/extras/hooks/decorator.rs
xavierforge: It is not opt-in to anything wasm related: `rig::tool::ToolDyn` (the trait `HookedTool` must implement to wrap tools for dynamic dispatch) declares `definition` and `call` as returning `WasmBoxedFuture`, so the impl signature has to match. On native targets the alias is just `Pin<Box<dyn Future + Send>>`, an ordinary boxed future; rig only drops the Send bound when compiling for wasm32. decorator.rs is deliberately the only rig-typed file in the hook system, and this type comes with that one contact surface. Added a comment in a24bb59.

src/agent/runner.rs
xavierforge: Leftover, and the setting is not lost. The parameter's only consumer was `.multi_turn(max_turns)`, which the manual loop replaced. That call was always a redundant self-override: rig seeds each request's turn bound from `agent.default_max_turns`, set at construction from the exact same `cli.resolve_max_agent_turns(cfg)` value the call sites passed, and `.multi_turn()` merely overrides that default with the identical number. The TUI path (`spawn_agent`) has always relied on the default alone. So the user-facing bound on tool-call round trips per prompt still applies unchanged; the new outer loop only handles `Stop` hook continuations and has its own separate cap (`MAX_STOP_BLOCKS = 8`). Removed the dead parameter from `run_print` and its callers in 19cb0f2.

src/agent/runner.rs
xavierforge: Done in 19cb0f2: replaced the tuple with a documented `LoopInfo { iteration, active }` struct (iteration is the 1-based --loop iteration, active is whether the loop continues after this one; they feed the `Stop` envelope's `loop_iteration`/`loop_active` fields). No change to the envelope JSON.

xavierforge:
While addressing the Windows comment I audited trust.rs and found a cross-project trust-replay gap: the trust hash covered only event + matcher + handler, and the store is user-level, so a binding like `{"command": "./guard.sh"}` trusted in one project would be auto-trusted in any other project shipping the identical definition, executing that project's file. Fixed in de7b626 by scoping the hash to the project root (subagent trust inheritance still works, same CWD). Also added a test proving a `PreToolUse updatedInput` rewrite cannot bypass permission deny rules (c05a90e). Note: the hash change invalidates any existing trusted-hooks.json entries, which is safe since the feature is unreleased.

gi-dellav:
Thank you for this big contribution, merged!

PR #185, opened 2026-07-12
feat(memory): add memory_edit, backups, and long-term append dedup

xavierforge:
## Summary

Adds `memory_edit`, a precise in-place edit and delete tool for the `memory` feature, makes destructive memory mutations recoverable via an automatic single-slot `.bak` backup, and stops long-term memory from accumulating duplicate lines on repeated append. This is PR-B of the memory v2 plan (PR-A, section-priority injection, already merged as `memory-context-injection`).

## Motivation

The memory write path previously supported only append and whole-file overwrite. Correcting or removing a single stale fact meant reading the entire long-term store, rewriting it in full, and overwriting: a high-risk operation where one hallucination during the rewrite can corrupt or erase unrelated memories, so in practice it never happened. Memory only grew, and stale or duplicate facts accumulated. This PR gives the model a surgical way to fix one line, a safe way to delete a stale note, a recoverable undo for destructive writes, and automatic suppression of duplicate long-term lines.

## Design decisions

Strict unique-substring matching, no fuzzy fallback. `memory_edit` requires `old_str` to match the target file exactly once (`match_indices` count == 1); zero or multiple matches is a hard error naming the count, with no write. Unlike the source-code `EditTool` (exact, then normalized, then Levenshtein), memory files are the model's own curated durable state, not code under review: a silent fuzzy match landing on the wrong line corrupts memory in a way that is easy to miss and costly to recover. The files are small plain Markdown the model can re-read verbatim, so exact matching is cheap and gives the simplest 0/1/N test matrix.

Note deletion lives in `memory_edit` (omit `old_str`), not a `memory_write mode=delete`. Adding a delete mode to `memory_write` would leave `content` a required-but-ignored field for that call. `memory_edit` is already the "remove or change memory" tool and already carries `name`, so no `memory_write` schema change is needed. Deletion is rejected for any non-note target, so long-term and scratchpad can never be fully deleted via this path.

Backup scope is deliberately narrow. A single-version `.bak` (overwriting any prior `.bak`, sibling to the original) is written before content-destroying operations only: long-term/scratchpad overwrite, long-term/scratchpad edit, and note deletion. Append is non-destructive by construction and never backs up; daily/note content edits are low-risk targeted replacements and are left un-backed-up to avoid churn. The `.bak` extension keeps backups out of `memory_read source=list` and `memory_search`, which both filter to `.md`. If the backup copy itself fails, the mutation still proceeds (the primary operation is what was asked for) but the tool response is suffixed with a warning so the caller knows there is no undo for that change.

Daily edits address a day by `name`. `memory_edit target=daily` honors a `name=YYYY-MM-DD` to edit an earlier day (defaulting to today), matching how `memory_read` already selects daily logs. The date is validated (ASCII digits and hyphens only) before it is spliced into a path, since `daily_file` does no sanitization of its own.

Platform support: this change is pure filesystem and string work (`std::fs`, `atomic_write`'s temp-file-then-rename, `Path::with_extension`), no process spawning or shell-outs, so it carries the same cross-platform support as the rest of the crate, including Windows. Paths are built with `Path`/`PathBuf` throughout; note names are sanitized by the existing `note_path` (rejects slashes and dots) and daily date names by the new digits-and-hyphens check before any file operation.

## Testing

- `cargo test --all-features`: 787 passing, 0 failing.
- New coverage in `src/tests/memory_tests.rs`: unique-match replace, 0/N-match rejection leaving the file byte-for-byte unchanged, empty-`new_str` substring deletion, note deletion by name, deletion rejected for non-note targets, missing-note error, daily edit targeting a named earlier day and rejecting an unsafe name, backup created/omitted across all target+operation combinations, single-version `.bak` overwrite, a backup-failure path that warns yet still completes the mutation, dedup against existing content and within a batch (including full-width U+3000 whitespace), all-duplicate batch leaving the file untouched, and non-long-term appends never deduplicating.
- The subagent-exclusion and backup-hidden-from-list tests drive the real production code paths (`subagents::builder::subagent_memory_tools()` and `Mem::list()`), not hand-copied mirrors, so a regression that leaked `memory_edit` into subagents or a `.bak` into the list would fail the suite.

## Reviewing

Seven commits, each a coherent step; most of the added lines are tests. The behavior change itself is small: `backup_file`/`backup_warning`/`normalize_line`/`is_safe_daily_name` helpers, the append-dedup block in `Mem::write`, `Mem::edit`, and the `MemoryEdit` rig tool (the last is boilerplate matching the sibling `MemoryWrite`/`MemoryRead` tools).

Suggested reading order:
1. `src/extras/memory/mod.rs`: `Mem::edit` (the new operation), then the `Append` branch of `Mem::write` (the dedup state machine), then the `backup_file`/`backup_warning`/`normalize_line`/`is_safe_daily_name`/`Mem::list` helpers.
2. `docs/MEMORY.md`: the prose contract for all of the above.
3. `src/agent/builder.rs`, `src/agent/prompt.rs`: one-line tool registration and prompt text.
4. Tests last.

Commit order: (1) unique-match content replacement, (2) whole-note deletion, (3) backup before destructive mutation, (4) long-term append dedup, then three review-hardening commits: (5) honor and validate a daily date in `memory_edit`, (6) warn when a backup cannot be written, (7) validate the daily date name in the `memory_read` path.

## Notes

- `src/extras/subagents/builder.rs` is modified even though it registers no new tool for subagents. The edit extracts the read-only memory tool set into a `subagent_memory_tools()` function that the builder now uses, so the "subagents cannot mutate memory" guarantee has a single source of truth a test can assert against directly. The guarantee itself is unchanged (subagents still get only `memory_read` and `memory_search`).
- Commit (7) fixes a pre-existing traversal in `memory_read`: its `daily` source spliced the caller-supplied `name` straight into `daily_file` with no sanitization. It is included here because this PR introduces the shared `is_safe_daily_name` check that closes the same class for `memory_edit`; reusing it for the read path is a one-line change.
- `memory_edit` deletes a whole note when `old_str` is omitted (with `target=note` and a `name`); this is intentional. `new_str` is ignored on that path, so a caller that means "overwrite a note" should use `memory_write mode=overwrite`, not `memory_edit`.
- Whole-line deletion via `memory_edit` requires including the trailing newline in `old_str`; otherwise a blank line is left behind. This is documented in the tool description and is self-correctable with a follow-up edit.
- Unrelated to this change: `just check` currently reports pre-existing `cargo fmt`/`clippy` failures in `src/setup/mod.rs` (identical to `main`; the project's CI runs `cargo build` + `cargo test`, not the local lint gate). This PR touches none of that file, and its own diff is `cargo fmt` and `clippy -D warnings` clean.

### DIFF
--- a/docs/MEMORY.md
+++ b/docs/MEMORY.md
@@ -120,7 +120,7 @@ Rules:
 
 ## Rig Tools
 
-Three tools are registered when the `memory` feature is enabled:
+Four tools are registered when the `memory` feature is enabled:
 
 ### `memory_write`
 
@@ -139,6 +139,19 @@ Three tools are registered when the `memory` feature is enabled:
 | `name` | string (opt) | Note stem or YYYY-MM-DD for daily |
 
 `source=list` enumerates all `.md` files in the store (global MEMORY.md + current project's notes + daily logs).
+
+### `memory_edit`
+
+| Parameter | Type | Description |
+|---|---|---|
+| `target` | string | `long_term`, `scratchpad`, `daily`, or `note` |
+| `name` | string (opt) | Note stem, required for `note` |
+| `old_str` | string (opt) | Substring to replace; must occur exactly once. Omit to delete a whole note |
+| `new_str` | string | Replacement text; empty string deletes the matched substring |
+
+Replaces a unique substring in a memory file in place. `old_str` is matched literally (no fuzzy matching) and must occur exactly once in the target file; zero or multiple matches fail without writing. `new_str` replaces the match verbatim with no newline cleanup, so an empty `new_str` deletes exactly the matched text, and including the trailing newline in `old_str` deletes a whole line.
+
+Omitting `old_str` deletes an entire note file (`notes/<name>.md`) from disk; this requires `target=note` with a `name`. Omitting `old_str` for `long_term`, `scratchpad`, or `daily` is rejected and changes nothing. Deleting a note that does not exist is an error.
 
 ### `memory_search`
 
--- a/docs/MEMORY.md
+++ b/docs/MEMORY.md
@@ -145,7 +145,7 @@ Four tools are registered when the `memory` feature is enabled:
 | Parameter | Type | Description |
 |---|---|---|
 | `target` | string | `long_term`, `scratchpad`, `daily`, or `note` |
-| `name` | string (opt) | Note stem, required for `note` |
+| `name` | string (opt) | Note stem (required for `note`); or a `YYYY-MM-DD` date for `daily` to edit an earlier day (defaults to today) |
 | `old_str` | string (opt) | Substring to replace; must occur exactly once. Omit to delete a whole note |
 | `new_str` | string | Replacement text; empty string deletes the matched substring |
 
--- a/docs/MEMORY.md
+++ b/docs/MEMORY.md
@@ -170,6 +170,8 @@ A backup is taken only before these operations (and only when the target file al
 
 Appends are non-destructive by construction, so they never back up. `daily` and `note` content edits are targeted unique-match replacements (low-risk and already reversible via a re-edit), so they are deliberately left un-backed-up to avoid churn.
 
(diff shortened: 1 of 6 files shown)


gi-dellav:
Thanks, merged

PR #200, opened 2026-07-19
feat(ui): sort prompt list alphabetically and show each prompt's source

wowi42:
## Problem

`/prompt` listed prompts in HashMap iteration order in the pickers, with no hint about where each prompt came from — which matters since prompts can live in up to four places (built-in, global data dir, `data/prompts/`, `.zerostack/prompts/`, later overriding earlier).

## Changes

- **Sorted alphabetically** — the `/prompt` picker, the `.` dot picker, and the `/prompt` command-picker branch now list prompts sorted (the bare `/prompt` text listing already was).
- **Source tags** — every entry shows its provenance in parentheses: `code (built-in)`, `myprompt (global)`, `x (local)`. The two project-level sources (`data/prompts/`, `.zerostack/prompts/`) both display as `local`.

Implementation:

- `context::prompts`: new `PromptSource` enum + `load_with_sources()` tracking which highest-priority location each prompt came from; replaces the old `load()` (no remaining callers). Stored as `ContextFiles.prompt_sources`; `/regen-prompts` refreshes it and the picker state.
- `ListPicker` gained optional per-item **annotations**: drawn as `name (source)` but never part of filtering or selection — typing `local` won't match every local prompt, and selecting an entry still runs `/prompt <name>` with the bare name.
- Docs: `COMMANDS.md` `/prompt` entry updated; `CONFIG.md` source lists corrected to the actual post-`214a1b3` paths (`data/prompts/`, `data/themes/`, `$ZS_DATA_DIR` defaults).

## Verification

- `cargo fmt --check` ✅
- `cargo test --locked` ✅ (610 passed, incl. 7 new: 6 provenance + 1 picker-annotation)
- `cargo test --locked --no-default-features` ✅ (512 passed)
- `cargo test --locked --all-features` ✅ (814 passed)
- `cargo clippy --locked --no-default-features -- -D warnings` ✅
- `cargo clippy --locked --all-features -- -D warnings` ✅
- `cargo install --locked --path . --debug` ✅

### DIFF
--- a/docs/COMMANDS.md
+++ b/docs/COMMANDS.md
@@ -93,7 +93,7 @@ starting a session or making a model call. See
 
 | Command | Description |
 | ------- | ----------- |
-| `/prompt` | List available prompts. |
+| `/prompt` | List available prompts, sorted alphabetically, each tagged with its source (`built-in`, `global`, `local`). The `/prompt` and `.` pickers show the same tags. |
 | `/prompt <name>` | Activate a named prompt. Also applies `%%mode=` from the prompt file if present (see below). |
 | `/prompt default` | Clear the active prompt. |
 
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -17,14 +17,14 @@ overriding earlier ones for same-named files:
 
 **Prompts** (priority low to high):
 1. Embedded at compile time
-2. `~/.local/share/zerostack/prompts/` (global, user-level)
-3. `prompts/` (project-local, relative to CWD)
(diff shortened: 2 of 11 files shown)


gi-dellav:
I don't think that this is a needed change, as prompt are almost always called with the first couple of letters (ex. `.c`+tab = `.code`; users often don't scroll between prompt, but just use the first chars until the prompt is selected), and often the source is not relevant (and if the user in interested in finding it, they can just follow the load order and see where are the prompts located).

PR #209, opened 2026-07-23
test: serialize track_read tests sharing global state

wowi42:
## Problem

`track_read_duplicate_returns_blocking_message` (and its neighbors in `src/tests/tools_mod_tests.rs`) fails intermittently under parallel test execution.

Root cause: the 9 tests in that file share two process-wide globals — the `DENY_REPEATED_READS` flag and `READ_TRACKER` (`src/agent/tools/mod.rs`). One test (`track_read_returns_none_when_deny_disabled`) sets the flag to `false`; if that lands between another test's `set(true)` and its second `track_read` call, the "duplicate" read returns `None` and the assertion fails. Observed in practice while running the suite on #208.

## Fix

Add a file-local `TRACK_TEST_LOCK: Mutex<()>` and hold the guard for the whole body of each test that touches these globals, so flag flips and tracker mutations can't interleave. Each test already sets the flag it needs at the start, so no ordering dependency remains. The pure `skip_*` tests don't take the lock.

## Verification

- Full suite: 648 passed, 0 failed — 5 consecutive runs
- Targeted: `cargo test tools_mod_tests` — 10/10 runs green

### DIFF
--- a/src/tests/tools_mod_tests.rs
+++ b/src/tests/tools_mod_tests.rs
@@ -1,6 +1,16 @@
 use crate::agent::tools::{
     deny_repeated_reads, is_skip_dir, set_deny_repeated_reads, track_read, untrack_read_path,
 };
+use std::sync::{Mutex, MutexGuard};
+
+/// Serializes the tests that touch the global deny-repeated-reads flag and
+/// read tracker: they share process-wide state and would otherwise race
+/// under parallel test execution.
+static TRACK_TEST_LOCK: Mutex<()> = Mutex::new(());
+
+fn track_test_lock() -> MutexGuard<'static, ()> {
+    TRACK_TEST_LOCK.lock().unwrap_or_else(|e| e.into_inner())
+}
 
 #[test]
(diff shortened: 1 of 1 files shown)


wowi42:
Hi @gi-dellav — the failing `package` check on this PR is a transient network error, not a code issue. The job failed while downloading a dependency from crates.io:

```
error: failed to get `agent-client-protocol-derive` as a dependency of package `agent-client-protocol v1.3.0`
  download of ag/en/agent-client-protocol-derive failed
  curl failed: [16] Error in the HTTP2 framing layer
```

All other jobs (fmt, clippy, tests on ubuntu + macOS with all feature combos) pass. Could you hit "Re-run failed jobs" on the workflow run when you get a chance? I don't have the permissions to re-trigger it myself. Thanks!

gi-dellav:
No problems, I see that this is just a CI issues (something similar sometime happens with the GitHub pages CI), so I'll just merge it, the code looks fine

PR #215, opened 2026-07-24
fix(prompts): correct stale references and mode contradictions

wowi42:
First of two PRs from a full review of the 16 built-in prompts in `data/prompts/`, with every reference verified against the implementation. This one is the quick-fix sweep; a structural PR (orchestrator rewrite, work.md trimming, boilerplate dedup) can follow.

## Fixes

**Wrong references**
- Replace the stale `'oldString not found'` error reference (a Claude Code-ism) with the edit tool's real error text `'search text not found'` (`src/agent/tools/edit.rs:306`) — 8 files: code, debug, default, autoconfig, frontend-design, refactor, simplify, write-text (and write-prompt below)
- `default.md`: sibling prompt list now includes `work`, `orchestrator`, `write-text` (it listed only 12 of 15)
- `work.md`: all 5 cron examples used the nonexistent `--custom-prompt=work` flag → `--load-prompt work` (`src/cli.rs:19`)
- `write-prompt.md`: its job is authoring prompts, yet it never documented the required first-line `%%mode=X` directive, the valid modes, or the 4-layer override locations — added a `zerostack Prompt Format` section; also clarified that "superpower / claude-plugins" are third-party skill formats
- `autoconfig.md`: documents the `.zerostack/prompts` and `.zerostack/config.toml` project-local layers and platform data-dir paths; `external_directory` correctly described as a top-level map, not a per-tool rule

**Mode contradictions (bash is auto-denied in readonly/planwrite)**
- Removed dead bash guidance from `ask`, `brainstorm`, `plan`, `review`, `review-security` ("use bash directly for VCS", chain-with-`&&`, quoting bullets)
- `review.md` was the worst case — its primary input is a diff it couldn't fetch; it now explicitly says bash is denied in readonly and to review files directly or ask the user for the diff
- `ask.md`: added `list_dir` to the permitted tools (allowed in readonly), removed a duplicated rule and unreachable commit/push safety bullets

**Contradictions & fabrications**
- `review-security.md`: unified three conflicting confidence rules into one policy (HIGH → report, MEDIUM → "Needs verification", LOW → don't report)
- `refactor.md`: "improve error handling" no longer contradicts "don't change error semantics" (scoped to internal robustness vs public semantics), and the process now runs a baseline test suite *before* refactoring (the "STOP on pre-existing failures" rule was unactionable without it)
- Removed the fabricated `~50 lines per edit` "tool limit" (frontend-design, code, refactor, simplify — no such limit exists)
- `work.md`: the "act autonomously" principle now states the boundary — local files/commands are routine; anything reaching external services still requires confirmation

All `%%mode=` directives were audited and are valid and well-chosen.

## Verification

- `grep` sweep confirms no stale strings remain
- `cargo test`: 687 passed, 0 failed (prompts are embedded via `include_dir!`)
- `cargo install --path . --debug` succeeds

### DIFF
--- a/data/prompts/ask.md
+++ b/data/prompts/ask.md
@@ -42,10 +42,6 @@ When web search MCP tools (Exa, Context7, Grep.app) are available:
 - Combine related queries into a single batch of parallel calls.
 - Prefer official documentation sources over community answers.
 
-## Safety Rules
-
-- Do not execute shell commands that modify the user's system outside the workspace without asking.
-
 ## Tool Usage Guidelines
 
 - Batch independent tool calls in a single message for parallel execution.
--- a/data/prompts/ask.md
+++ b/data/prompts/ask.md
@@ -1,6 +1,6 @@
 %%mode=readonly
 
-You MUST NOT use write, edit, or bash. Only read, grep, and find_files are permitted.
+You MUST NOT use write, edit, or bash. Only read, grep, find_files, and list_dir are permitted.
 
 If the user asks for changes, tell them to switch to a coding prompt (code, debug, or default).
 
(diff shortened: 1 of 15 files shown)


gi-dellav:
Looks nice, but why in some files the security rules are reduced? For example, `data/prompts/ask.md`

wowi42:
The rules were removed only in ask.md and brainstorm.md. Both run in %%mode=readonly. In that mode the permission system blocks write, edit, and bash, so the agent cannot run git commands at all. The VCS rules were dead text there. Every prompt where bash is allowed (code, debug, default, plan, and the others) keeps the rules unchanged.

But you are right that it is not consistent. ask.md still has a shell rule that readonly mode makes useless too, and review.md and review-security.md are also readonly but kept the VCS rules.

I will finish the cleanup: all readonly prompts drop the rules that only apply to changing code or git. Enforcement lives in the permission layer, not in the prompt text. I will push a commit for this.

wowi42:
Done in 7a04b66. ask.md and brainstorm.md lost the leftover shell rule, review.md and review-security.md lost the VCS and secrets rules. All readonly prompts are now clean, and all coding prompts keep the full safety rules.

PR #228, opened 2026-07-27
feat(session): record active prompt name and source

### LINKED ISSUE #227: Session files do not record which prompt was loaded
## Current behavior

The name of the prompt a session is running under lives only in memory, on `ContextFiles::current_prompt_name` (`src/context/mod.rs:61`). It is set at startup from `default_prompt` (`src/startup.rs:626-661`, assignment at `:651`) or from `--load-prompt` (`src/startup.rs:663-706`, assignment at `:687`), and updated when the prompt changes at runtime, mostly through `apply_prompt_mode` (`src/ui/mod.rs:62-78`, assignment at `:76`).

It surfaces in exactly one place: the status bar item (`src/ui/statusline.rs:263`). It is not written to the session file, `Session` (`src/session/mod.rs:61-124`) has no field for it, and it is not logged either.

So a session file on disk does not say whether the run used `code`, `plan`, `review`, one of the other built-ins shipped under `data/prompts/` (16 of them today), or a user prompt from the global prompts dir, `data/prompts/`, or `.zerostack/prompts/`, any of which can also shadow a built-in name with different content (`src/context/prompts.rs`, `load()`).

## Why this matters

Since 4b01639 ("fix: apply the default prompt's `%%mode=` directive at startup") a prompt's `%%mode=` directive actually takes effect at the permission layer. Which prompt was loaded now changes what tools the run was allowed to do. With that fact absent from the session file, two sessions with the same model, same working directory, and same messages can have had different permission behavior, and nothing in the file explains why. That makes a saved session hard to reproduce and hard to debug after the fact.

Automation that reads session JSON has the same gap: it can see the conversation and the model, but not the prompt that shaped it.

## Proposal

Record the active prompt in the session file. Deliberately minimal:

- Add `#[serde(default)] pub prompt: Option<PromptRef>` to `Session`, where `PromptRef` is `{ name, source }` and `source` distinguishes a built-in (embedded under `data/prompts/`) from a user file. `context::prompts::load()` already walks the layers in priority order but returns a flat `HashMap<String, String>`, so the origin is discarded today and would need to be threaded through or recomputed.
- Write it during startup prompt resolution, in both the `default_prompt` block (`src/startup.rs:626-661`) and the `--load-prompt` block (`src/startup.rs:663-706`).
- Keep it in sync when the prompt changes at runtime. `apply_prompt_mode` (`src/ui/mod.rs:62-78`) is the single choke point for `/prompt <name>` (`src/ui/slash/content.rs:65`), the `.<prompt>` dot prefix (`src/ui/app.rs:1069`, `:1090`), `/review` (`src/ui/slash/review.rs:62`), and chain phase transitions (`src/ui/app.rs:1015`). Three sites assign `current_prompt_name` directly and need the same treatment: `/prompt default` clearing it (`src/ui/slash/content.rs:41-42`) and the two dot prefix restore paths (`src/ui/app.rs:903`, `:972`).
- `#[serde(default)]` keeps existing session files loading.

One implementation note: `apply_prompt_mode` takes `&mut ContextFiles` and does not have the `Session` in hand, so either its callers write the field or the signature grows a session argument. Either is small.

No history of switches, no snapshot of prompt contents: last one wins, matching what the status bar already shows.

Happy to send this as a PR.

xavierforge:
Closes #227

## Current behavior

The prompt shaping a session lives only in memory, on `ContextFiles::current_prompt_name`. It is set at startup from `default_prompt` or `--load-prompt`, and updated at runtime through `apply_prompt_mode` plus a few direct assignment sites. It surfaces in exactly one place, the status bar. The session file has no field for it, so a saved session cannot say whether it ran under `code`, `plan`, `review`, or a user prompt file, even though a prompt's `%%mode=` directive now drives permission enforcement.

## What changes

- `Session` gains `#[serde(default)] pub prompt: Option<PromptRef>`, where `PromptRef` is `{ name, source }` and `source` is `BuiltIn` or `UserFile`.
- `context::prompts` gains `source_of(name)`, a small helper that checks the same override directories `load()` already walks (global prompts dir, the project's `data/prompts/`, `.zerostack/prompts/`) to classify a resolved prompt's origin, without changing `load()`'s `HashMap<String, String>` return shape.
- Startup prompt resolution writes the field in both the `default_prompt` block and the `--load-prompt` block.
- `apply_prompt_mode` now takes the session and writes the field, so its five call sites (`/prompt <name>`, `/review`, chain phase transitions, both dot-prefix branches) pick it up by construction. The three sites that assign `current_prompt_name` directly instead of going through `apply_prompt_mode` (`/prompt default`, and the two dot-prefix restore paths) get the same treatment.
- No switch history is kept; the last prompt active wins, matching what the status bar already shows.
- `#[serde(default)]` keeps existing session files loading with `prompt: None`.

## Tests

- `src/session/mod.rs` schema plus a save/load round trip test and a test that a session file written before this field existed still loads, with `prompt: None`.
- `src/context/prompts.rs`: `source_of` unit tests for a built-in prompt, a user file shadowing a built-in name, and a prompt that only exists as a user file.
- `src/tests/prompt_mode_tests.rs`: `apply_prompt_mode` now asserts the session's recorded prompt name and source, including that an unknown prompt name is a no-op that leaves the previous record untouched, and that switching twice leaves the record reflecting the prompt active last.

Verified with `cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings`, both clean.

### DIFF
--- a/src/context/prompts.rs
+++ b/src/context/prompts.rs
@@ -32,6 +32,49 @@ pub fn load() -> HashMap<String, String> {
     prompts
 }
 
+/// Whether prompt `name`'s currently-loaded content is the compiled-in
+/// default or a user customization.
+///
+/// Compares content rather than testing file existence: `ensure_global()`
+/// seeds every embedded prompt into the global prompts dir on first run, so
+/// `code.md` being present on disk says nothing about whether the user
+/// touched it. A prompt counts as [`UserFile`](crate::session::PromptSource::UserFile)
+/// only when the text that actually shaped the session differs from the
+/// compiled-in one, or has no compiled-in counterpart at all.
+///
+/// Reads the highest-precedence on-disk copy, mirroring the last-writer-wins
+/// order in [`load()`]: `.zerostack/prompts/`, then the project's
+/// `data/prompts/`, then the global dir. A lower-precedence copy being edited
+/// is irrelevant when a higher-precedence one shadows it.
+pub fn source_of(name: &str) -> crate::session::PromptSource {
+    use crate::session::PromptSource;
+
+    let file_name = format!("{name}.md");
+    let effective = [zerostack_dir(), PathBuf::from("data/prompts"), global_dir()]
+        .into_iter()
+        .find_map(|dir| std::fs::read_to_string(dir.join(&file_name)).ok());
+
+    match effective {
+        // Nothing on disk, so `load()` served the embedded copy. A name with
+        // no embedded copy either cannot reach here: callers resolve against
+        // `load()`'s map first.
+        None => PromptSource::BuiltIn,
+        // On disk but byte-identical to the embedded default: a seeded copy,
+        // not a customization. Also covers a name with no embedded
+        // counterpart, which can only have come from a user file.
+        Some(content) => {
+            let embedded = EMBEDDED
+                .get_file(&file_name)
+                .and_then(|f| f.contents_utf8());
+            if embedded == Some(content.as_str()) {
+                PromptSource::BuiltIn
+            } else {
+                PromptSource::UserFile
+            }
+        }
+    }
+}
+
 pub fn ensure_global() -> anyhow::Result<()> {
     let dir = global_dir();
     if !dir.exists() {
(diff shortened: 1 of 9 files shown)


gi-dellav:
Please fix conflicts; for the rest, it's ready to be merged

xavierforge:
Conflicts fixed (rebased onto current `main`; the only overlap was `Session` gaining fields from both #230 and this PR).

While rebasing I found a defect in this PR worth fixing before merge, so the force-push also changes behavior. `source_of()` decided built-in vs user-file by testing whether `<name>.md` existed in any of the three prompt dirs. But `context::load()` calls `prompts::ensure_global()`, which seeds every embedded prompt into the global prompts dir on first run, so `code.md` always exists there and the field reported `user_file` for every prompt: `BuiltIn` was unreachable in production. The unit tests missed it because they point `ZS_DATA_DIR` at an empty temp dir and never call `ensure_global()`.

It now compares content instead: read the highest-precedence on-disk copy (mirroring `load()`'s override order) and report `UserFile` only when that text differs from the compiled-in one, or has no compiled-in counterpart. Added three tests, one of which calls `ensure_global()` first and pins `code` as `BuiltIn` (it fails against the old logic).

PR #231, opened 2026-07-27
feat(subagents): record subagent tool calls in headless sessions

### LINKED ISSUE #229: Headless `-p` runs record no tool calls in the session file
## Current behavior

A headless run (`zerostack -p "..."`) writes exactly two messages to its session file: the user prompt and the final assistant text. `dispatch_print` (`src/startup.rs:787`) calls `add_message` twice and nothing else (`src/startup.rs:879-880`). Every tool the model ran during that turn is absent from the saved session.

The only trace a tool call leaves is on stdout, and only when `--pure-stdout` is passed: `run_print` prints a `◈ <name> <args summary>` line (`src/agent/runner.rs:649`) and a `◈ <name> result:` block (`src/agent/runner.rs:671`) capped at 40 lines. Without that flag there is no trace at all, and with it the trace is terminal output, not part of the session.

The TUI does not have this gap. `AgentEvent::ToolCall` / `SubagentToolCall` / `ToolResult` are persisted via `Session::add_tool_call` / `add_subagent_tool_call` / `add_tool_result` (`src/session/mod.rs:321-354`), called from `src/ui/event_handler.rs:109`, `:119`, and `:128`. The same turn run in the TUI produces a session containing `tool_call` and `tool_result` messages; run with `-p`, it does not.

Two things about the TUI path are worth stating, since a fix should not inherit them silently:

- The `tool_call` message stores only the display summary produced by `format_tool_call_summary` (`src/ui/utils.rs:141-180`, plus `format_task_summary` at `:182-196` for the `task` tool). That helper keeps one or two representative argument values, truncates each to 200 chars (`TOOL_SUMMARY_MAX`, `src/ui/utils.rs:6`, applied in `display_value`, `src/ui/utils.rs:8-14`), and for `bash` splices the raw command in without quoting. An actual saved message looks like this:

  ```json
  {"role": "tool_call", "content": "bash grep -r \"HOOK_DENY_ME\" . 2>/dev/null || echo \"Not found\"", "estimated_tokens": 15}
  ```

  There is no structured record of what the tool was actually called with, and no reliable way to tell where the tool name ends and the arguments begin.

- `convert_history` (`src/agent/runner.rs:163-183`) folds `ToolCall` / `ToolResult` / `SubagentToolCall` messages back into replayed history. So when a TUI session is resumed the model sees what earlier tools did, and when a headless session is resumed with `-p --continue` it does not, because those messages were never written. This is separate from #212: even with prior history correctly passed to `run_print`, the tool context simply is not in the file to pass.

## Why this matters

- The session file is meant to be the record of what actually happened in a turn. Under `-p` that record is incomplete, so debugging a headless run means going back to terminal output that may not have been kept, and that was capped at 40 lines per result anyway.
- Anything that reads session JSON (scripts, automation, `/export`) gets the conversation but none of the tool-level facts.
- The same turn gets different resume behavior depending on whether it ran in the TUI or headless, which is a parity gap rather than a design choice.

## Proposal

Add an optional structured `tool` field to `SessionMessage`, and populate it from both paths.

1. **Schema.** `SessionMessage` (`src/session/mod.rs:26-30`) gains `#[serde(default)] pub tool: Option<ToolRecord>`. Existing session files keep loading unchanged, the same way the `#[serde(default)]` fields already on `Session` do (`src/session/mod.rs:68-91`).

2. **`tool_call` messages** carry `tool: { id, name, args }`, with `args` holding the complete JSON arguments. This is the part that is genuinely lossy today. `content` keeps the existing summary, so the TUI transcript, `/export`, and token estimation all render exactly as they do now.

3. **`tool_result` messages** carry `tool: { call_id, name, truncated, full_output_path? }`. The result text itself is not duplicated into the new field: it stays in `content` under the rules that already exist, full text under 12,000 chars (`TOOL_RESULT_SAVE_THRESHOLD`, `src/session/mod.rs:10`), otherwise the current overflow file plus head/tail plus path (`tool_result_content`, `src/session/mod.rs:334-346`, and `format_truncated_tool_result`, `src/session/mod.rs:589`). No new truncation policy is introduced.

4. **`id` / `call_id`** are a per-session counter, so a result links back to its call explicitly instead of by position. Adjacency is not good enough: subagent events land between an outer `task` call and its result, so positional pairing mispairs.

5. **Where the code goes.** The new fields are filled in inside the shared `Session` methods, so the TUI path picks them up by construction. The headless fix is then wiring: `run_print` (`src/agent/runner.rs:579`) has no event channel at all today, so it needs a way to hand tool call and result events back to `dispatch_print`, which already owns the `Session`.

6. **Subagents.** `AgentEvent::SubagentToolCall` already exists (`src/event.rs:15-19`) and is emitted from `run_subagent` (`src/agent/runner.rs:828-834`), but only the TUI and ACP consume it today. Route it into the headless session as well, stored with a `parent_call_id` pointing at the enclosing `task` call.

### Example

Two messages from a headless session after the change:

```json
{
  "role": "tool_call",
  "content": "bash cargo test --lib session",
  "estimated_tokens": 8,
  "tool": {
    "id": 4,
    "name": "bash",
    "args": { "command": "cargo test --lib session", "timeout": 120 }
  }
},
{
  "role": "tool_result",
  "content": "bash:\ntest result: ok. 214 passed; 0 failed",
  "estimated_tokens": 12,
  "tool": { "call_id": 4, "name": "bash", "truncated": false }
}
```

## Scope, and what this does not cover

- **Subagent tool results are out of scope.** There is no event for them: `run_subagent` forwards `StreamedAssistantContent::ToolCall` and drops the matching `ToolResult` (`src/agent/runner.rs:818-845`). Recording subagent results would mean adding a new event first, which belongs in its own change.
- **`--loop` is not covered.** The loop path calls `run_print` with empty history and carries no `Session` at all (see `TODO(loop-history)`, `src/extras/loop/headless.rs:67`). Making `--loop` write tool records depends on the loop owning a real `Session`, which is a separate change.
- **One deliberate behavior change.** Once tool messages exist in headless sessions, `-p --continue` will replay tool interactions into the model's history, because `convert_history` already handles those roles. That is the point: it makes headless resume behave like TUI resume. It does mean resumed headless turns carry more context than they do today.

## Offer

Happy to implement this. Suggested split into two PRs:

1. The `tool` field on `SessionMessage` plus main-agent wiring for both the TUI and headless paths.
2. Subagent visibility in headless sessions, behind the existing `subagents` feature.

xavierforge:
Closes #229

Stacked on #230: until that one merges this branch also carries its two commits, so only the top commit (`feat(subagents): record subagent tool calls in headless sessions`) is new here and only that one needs review.

## What this adds

#229 has two halves. #230 covers the main agent's own tool calls and results in `-p` sessions; this covers the tool calls made by the subagents a `task` call spawns, which headless runs never recorded at all.

The events already exist: `run_subagent` emits an `AgentEvent::SubagentToolCall` per subagent tool call, and the TUI consumes them. Only `spawn_agent` ever published a sender for that channel, so under `-p` the sender was `None` and every subagent tool call was dropped. `run_print` now publishes one for the duration of the turn.

## Attribution

Subagent events are concurrent with the main agent's stream: they arrive between a `task` call and its result, not as stream items. The loop therefore waits on both at once through a biased `select!`, so anything already queued is consumed before the next stream item is handled. That gives each subagent call the main-agent call that was in flight when it arrived, which becomes `parent_call_id` once `dispatch_print` has stamped that call's id.

The same `select!` is what keeps the run alive. The event channel is bounded (32, the same capacity `spawn_agent` uses), and while a `task` call runs the consumer is parked inside the very tool the subagent's sender is feeding. With no live drain, a subagent making more tool calls than the channel holds blocks its sender forever and the turn never finishes. `subagent_calls_beyond_the_channel_capacity_do_not_stall_the_turn` pushes 200 calls through that channel and fails on its timeout if the drain is taken away. The `ToolResult` arm additionally sweeps whatever is still queued, which covers a tool that emits its events without ever yielding.

Nothing about stdout changes: the drain prints nothing, and `--pure-stdout` output is exactly what it was.

## Schema

`ToolRecord` (added in #230) gains a third `#[serde(untagged)]` variant:

```json
{ "parent_call_id": 4, "name": "grep", "args": { "pattern": "fn main" } }
```

`parent_call_id` is required and there is deliberately no `id` of its own: subagent tool results have no event to record, so nothing ever links back to one, and an `id` field would make this JSON satisfy the `Call` variant too, which is listed first and would win the untagged match. A round trip test pins both directions.

`Session::add_subagent_tool_call` takes the parent id as an `Option` and falls back to the previous behavior (display summary, no structured record) when there is none, instead of inventing a link. The TUI passes its `pending_tool_call_id` by peek, since the enclosing call's `ToolResult` still owns the consuming `take`. Both paths write through that one session method, so the TUI records gain the same structure by construction.

Old session files are unaffected: `SessionMessage.tool` is `Option<ToolRecord>` with `#[serde(default)]` from #230, and this only adds a shape that never existed before.

## Scope

Gated behind the existing `subagents` feature. Subagent tool results stay out of scope, since no event carries them today. `--loop` still records nothing, as that path has no `Session`.

## Lines

Top commit only, 597 added and 53 removed:

| Part | Added | Removed |
|---|---|---|
| Logic (`runner.rs`, `session/mod.rs`, `startup.rs`, `event_handler.rs`) | 155 | 11 |
| of which comments | 73 | |
| Tests (`src/tests/`) | 441 | 41 |
| Docs | 1 | 1 |

The test lines are 302 for the new headless e2e file, 94 for the session record cases, and the remainder for renaming the shared `run_print` test guard: `run_print` now reaches two process global singletons instead of one, so every `run_print` test has to serialize against the others, including in builds without the `hooks` feature.

## Verification

`cargo test --features subagents` (703 pass), `cargo test --all-features` (920 pass), `cargo fmt --check`, and `cargo clippy --all-targets --all-features -- -D warnings`.

### DIFF
--- a/docs/SUBAGENTS.md
+++ b/docs/SUBAGENTS.md
@@ -161,7 +161,7 @@ Key files:
 | `src/extras/subagents/task_tool.rs`          | `TaskTool` implementation             |
 | `src/extras/subagents/builder.rs`            | Subagent construction (`build_explore_agent`) |
 | `src/extras/subagents/prompt.rs`             | Subagent system prompt                |
-| `src/agent/runner.rs` (`run_subagent`)       | Silent agent execution                |
+| `src/agent/runner.rs` (`run_subagent`)       | Silent agent execution; reports each subagent tool call as an event |
 | `src/agent/builder.rs`                       | Wires `TaskTool` into main agent      |
 | `src/provider.rs` (`AnyAgent::run_subagent`) | Type-erased dispatch                  |
 | `src/main.rs`                                | Initializes `SubagentConfig`          |
--- a/src/agent/runner.rs
+++ b/src/agent/runner.rs
@@ -615,6 +615,27 @@ where
     // returned to the caller (`dispatch_print`) for session persistence. See
     // design.md decision 5.
     let mut recorded_interactions: Vec<ToolInteraction> = Vec::new();
+    // Subagent tool calls reach this loop on a side channel rather than as
+    // stream items: `run_subagent` sends them from the tokio task the `task`
+    // tool spawned, concurrently with this turn's own stream. Only
+    // `spawn_agent` (the TUI) ever published a sender, so headless runs left
+    // subagent activity untraced; publishing one here is what makes it
+    // recordable. Same channel capacity as `spawn_agent`'s, and drained by
+    // the `select!` below while the `task` tool is still running, so a
+    // subagent making many tool calls cannot fill it and stall.
+    #[cfg(feature = "subagents")]
+    let (subagent_tx, mut subagent_rx) = mpsc::channel::<AgentEvent>(32);
+    #[cfg(feature = "subagents")]
+    crate::extras::subagents::set_subagent_event_tx(subagent_tx.clone());
+    // Held for the whole turn: with no sender alive the channel would be
+    // closed, and the `select!` arm below would then complete immediately on
+    // every poll instead of waiting.
+    #[cfg(feature = "subagents")]
+    let _subagent_tx = subagent_tx;
+    // Subagent calls seen since the current main-agent tool call started;
+    // moved into that call's `ToolInteraction` when its result arrives.
+    #[cfg(feature = "subagents")]
+    let mut pending_subagent_calls: Vec<SubagentCall> = Vec::new();
     let mut usage = rig::completion::Usage::new();
     // Set true only when a `Stop` hook forces another turn; drives the outer
     // loop. Stays false (single pass, no continuation) in the hooks-off build.
@@ -630,7 +651,30 @@ where
 
     while continue_turn {
         continue_turn = false;
-        while let Some(item) = stream.next().await {
+        loop {
+            // Wait for the next stream item while staying available to the
+            // subagent channel. `StreamExt::next` is cancel-safe (it only
+            // polls the stream, which owns its own state), so losing the race
+            // to a subagent event costs nothing: the next iteration polls the
+            // same stream again.
+            #[cfg(feature = "subagents")]
+            let next_item = loop {
+                tokio::select! {
+                    // `biased`: drain everything already queued before
+                    // touching the stream, so a subagent event that arrived
+                    // during the `task` call is attributed to that call and
+                    // not to whatever comes next.
+                    biased;
+                    Some(event) = subagent_rx.recv() => {
+                        push_subagent_call(&mut pending_subagent_calls, event);
+                    }
+                    item = stream.next() => break item,
+                }
+            };
+            #[cfg(not(feature = "subagents"))]
+            let next_item = stream.next().await;
+
+            let Some(item) = next_item else { break };
             match item {
                 Ok(MultiTurnStreamItem::StreamAssistantItem(StreamedAssistantContent::Text(
                     text,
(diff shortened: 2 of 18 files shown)


xavierforge:
Folded into #230 so the tool record work reviews and merges as one unit. #230 now carries all three commits and closes #229.

PR #234, opened 2026-07-27
feat: show startup progress in the chat feed

wowi42:
## Summary

Launching zerostack dropped the user into an empty feed — no logo, no indication of what was loaded, and MCP servers connecting silently in the background. This PR makes the startup sequence report itself directly in the chat UI, where it stays as scrollable history (no separate splash screen, single alt-screen entry).

```
zerostack v1.7.2

Ready to code; type a request or '/' for commands

… Loading 3 MCP server(s)
✓ MCP Exa Web Search — 2 tools (577ms)
✓ MCP glitchtip — 5 tools (1.8s)
✓ MCP redmine — 12 tools (930ms)
✓ agent ready — 3 MCP server(s), 19 tools connected
```

## Changes

- **Banner**: plain `zerostack vX.Y.Z` line — no logo, in keeping with the IBM-style design. Model/context/prompt/branch/cwd are *not* repeated — they already live in the statusline.
- **Startup log** (new `src/ui/boot.rs`, pure + unit-tested), **opt-in via `show_additional_info_startup` (default `false`)**: global/local config, prompts/themes with project-local counts, context files, compile-time feature list, session restore, subagents, model info, tools & permissions — with per-step timings.
- **Live MCP loading**: servers connect in the open chat UI via `connect_all_with_progress`, one line per server with tool count and connect time; `McpClientManager` caches `tool_counts` and the agent-ready line totals them.
- **Surfaced warnings**: subagent provider fallback, advisor disabled, and missing sandbox backend were previously only in `tracing` logs — they now show as red `!` lines.
- **Bug fix**: `draw_bottom` panicked with u16 underflow ("attempt to subtract with overflow") on tiny/0-height terminals; input and cursor rows now reuse the already-saturating `input_top`.
- **Config**: new `show_additional_info_startup` key (default `false`) to opt into the startup info lines; MCP loading lines always show. Documented in `docs/CONFIG.md`.
- Headless paths (`-p`, `--loop`, `--acp`) are untouched — stdout stays clean.

## Verification

- `cargo test`: 706 passed (incl. new tests for boot state, banner, feed color override, durations, features list)
- `cargo fmt`, `cargo install --path . --debug`
- Smoke-tested on a real pty: single alt-screen entry, correct ordering/colors, timings, tool counts, no panic on 0×0 terminals


gi-dellav:
make sense, but:

1. I think that the gradient logo is a little bit excessive, break the IBM design that inspired the original zerostack TUI, and makes us the same as all other agents

2. While showing MCP is really useful, all other informations is useful only for a small subset of users, as it was never requested; I think it should be moved to an opt-in config flag called "show_additional_info_startup" or something similar

wowi42:
Thanks for the review — both points addressed in the merge commit:

1. **Logo removed.** The ASCII/gradient banner is gone; the banner is a plain `zerostack vX.Y.Z` line again, in keeping with the IBM-style design.
2. **Opt-in startup info.** All the non-MCP startup lines (config paths, prompts/themes, features, step timings, etc.) are now behind a new `show_additional_info_startup` config key, default `false`. Only the MCP loading lines (`… Loading N MCP server(s)`, `✓ MCP <name> — N tools (t)`, `✓ agent ready`) show by default, since those are the genuinely useful ones.

Also fixed the CI failures:
- `clippy --all-features`: boxed `PrebuildEvent::Done` (large_enum_variant), allowed the cfg-gated pushes in `active_features` (vec_init_then_push).
- `clippy`/`test --no-default-features`: irrefutable `if let` and `never_loop` in the no-mcp `resolve_prebuild`, unused `mut` / dead code under `--no-default-features`.

All four flavors now pass locally: `cargo clippy --all-features -- -D warnings`, `cargo clippy --no-default-features -- -D warnings`, `cargo test` (731), `cargo test --all-features` (952), `cargo test --no-default-features` (616).

Issue #186, opened 2026-07-13
TUI refactor checklist


wowi42:
## Goal
Make the TUI easier to maintain, test, and extend while keeping the current custom `crossterm` renderer. This issue is a checklist to tackle one item at a time.

## Architecture
- [x] Extract a stateful `App` struct from `run_interactive` in `src/ui/mod.rs:731-2809`. Keep the custom renderer; move event dispatch, agent lifecycle, and slash handling into methods.
- [x] Introduce a `Feed` model in `src/ui/feed.rs` that stores structured conversation blocks separately from `Renderer`. `Renderer` then draws `Feed::lines(width)`, making layout and scroll math testable without a terminal.
- [x] Group related state into structs: `AgentRunState`, `UiContext`, `ChainState`, `SlashState`. Replace the 10-20 parameter functions.

## Control flow and duplication
- [x] Replace magic error-string control flow (`DEFER_COMPRESS:`, `DEFER_INIT:`, etc.) with a typed `enum SlashOutcome { DeferCompress(...), DeferInit(...), ... }`. (taken by @gi-dellav)
- [x] Centralize agent rebuild in one `rebuild_agent(session, client, context, ...)` helper used by `start_main_run`, slash commands, and mid-turn compaction.
- [x] Extract a single `apply_prompt_mode(name, context, permission)` helper to replace the duplicated prompt-mode blocks.

## Rendering performance
- [x] Make `Renderer` dirty-region aware: track whether the chat buffer, input area, or statusline changed, and redraw only what changed.
- [x] Append streaming tokens to a running line and render the current unfinished line as plain text; parse markdown only for completed lines to avoid O(n²) re-parsing.
- [x] Pre-wrap the feed to viewport width so selection and scroll operate on visual rows instead of logical lines.

## Safety and error handling
- [x] Gate `!command` behind the same `PermissionChecker` used for tool calls, or remove the feature. (taken by @gi-dellav)
- [x] Fix `open_in_editor` to use `std::process::Command::new(editor).arg(path)` instead of shell interpolation, and extract one `suspend_tui` helper. (taken by @gi-dellav)
- [x] Surface non-fatal errors visibly instead of `let _ = ...` swallowing them.
- [x] Validate URLs in `open_url` and report clipboard/open failures to the status line.

## Input and pickers
- [x] Stream `FilePicker` results incrementally and cancel the walk task on `Esc`/deactivate.
[...]

PR #236, opened 2026-07-27
feat: character-level chat selection that persists after mouse release

wowi42:
## Summary

Implements item 4 of the "Input and pickers" section of #186 (*"Support character-level selection: mouse drag and `y` copy currently select whole visual rows only"*) and fixes a UX bug where the selection disappeared the moment the mouse button was released.

## Changes

**Character-level selection**
- Selection endpoints are now `(visual_line, display_col)` points instead of whole-row `LineEntry` indices; `MouseDrag` honors the column it previously ignored.
- The viewport renders the selection as a partial-line reversed segment: first/last lines sliced to their columns, middle lines fully covered. Slicing is display-column based and CJK-aware (wide chars straddling a boundary are excluded), via new `byte_index_at_display_col` / `display_col_slice` helpers in `ui::utils`.
- `selected_text()` (and therefore `y` copy) extracts exactly the highlighted range, normalizing reversed/upward drags.

**Selection persists after mouse release, and is copied on release**
- Previously `MouseUp` copied to clipboard and immediately cleared the selection, so the highlight vanished on release. Now the drag ends (`selection_dragging` tracks the button state separately from `selection_active`) but the highlight stays: `y` copies again, `Esc` clears, the next click starts a new selection.
- The selection is also copied to the system clipboard on release via `copy_to_clipboard` (pbcopy/wl-copy/xclip/clip.exe): while mouse capture is enabled the terminal emulator has no native selection, so Cmd+C (macOS) / Ctrl+Shift+C (Linux) has nothing to grab — the app must write the pasteboard itself.

## Verification

- `cargo test` (720 passed), `cargo test --no-default-features` (605), `cargo clippy --all-features` and `--no-default-features` with `-D warnings` — all clean.
- New unit tests: display-column helpers (incl. CJK), char-level `selected_text` (first/last line slicing, reversed drag, empty point selection).
- Smoke-tested on a pty with synthetic SGR mouse events: partial-line reverse highlight renders, selection survives release, `y` copies it.

Refs #186

### DIFF
--- a/src/tests/renderer_tests.rs
+++ b/src/tests/renderer_tests.rs
@@ -173,8 +173,8 @@ mod dirty {
         assert!(!r.chat_needs_redraw());
         // Selection fields are public and mutated directly by callers.
         r.selection_active = true;
-        r.selection_start = Some(0);
-        r.selection_end = Some(0);
+        r.selection_start = Some((0, 0));
+        r.selection_end = Some((0, 0));
         assert!(r.chat_needs_redraw());
         r.mark_chat_clean();
         r.clear_selection();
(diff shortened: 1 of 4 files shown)


wowi42:
Follow-up (0fa1d4d): selection is now also copied to the system clipboard on mouse release — with mouse capture enabled the terminal emulator has no native selection, so Cmd+C/Ctrl+Shift+C has nothing to copy and the app must write the pasteboard itself. The highlight still persists for y/Esc. Verified on macOS: pbpaste contains the selected text after release.

PR #253, opened 2026-08-08
fix(ui): restore cursor visibility on exit

thomasjfox:
Bring back the cursor when exiting zerostack. Noticed and tested the fix on "konsole".


gi-dellav:
Thanks for the PR!

You might want to join the Matrix chatroom, as you can find more development updates

thomasjfox:
> You might want to join the Matrix chatroom, as you can find more development updates

Hi Giuseppe! I'm not on Matrix (yet), but thanks for the invite!

PR #268, opened 2026-08-31
feat: add OrcaRouter as a built-in provider

dulcestentaciones2920-debug:
This PR makes [OrcaRouter](https://www.orcarouter.ai) selectable as a first-class provider in zerostack — the same way OpenRouter is today — so users of this minimal Rust coding agent can point `--provider orcarouter` at a single OpenAI-compatible gateway without hand-rolling a custom `base_url`. OrcaRouter is an OpenAI-compatible AI gateway built for both models and agents: like OpenRouter it exposes a provider/model namespace across many models, but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Adding `orcarouter` as a named provider means zerostack users get that whole stack directly, rather than treating OrcaRouter as an anonymous custom base URL. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

The implementation mirrors zerostack's existing OpenRouter wiring exactly: a new `ProviderKind::OrcaRouter`, `AnyClient`/`AnyModel`/`AnyAgent` variants that reuse rig's OpenRouter client pointed at `https://api.orcarouter.ai/v1`, its own `ORCAROUTER_API_KEY` env var, `orcarouter/auto` as the default model, and the same OpenRouter-style live `/models` pricing/context discovery plus `provider.order` routing for `anthropic/*` ids. It's also wired into the `/provider` picker, the `--setup` TUI, `--provider`/`ZS_PROVIDER`, `custom_providers` `provider_type`, subagents, the advisor, and the docs. OrcaRouter's `/models` listing reports OpenRouter-style `pricing`/`context_length`, so cost tracking and the context meter work from the first turn.

Verified locally: `cargo fmt --check`, `cargo clippy --all-features` and `--no-default-features` with `-D warnings`, and the full test suite all pass (the two `shell_mode_tests` kill-behavior cases fail identically on `main` in a container without process-group semantics). I also ran a live one-shot through the new code path — `zerostack --provider orcarouter --model orcarouter/auto -p 'Reply with pong'` against a real `ORCAROUTER_API_KEY` — and got a `pong` reply.

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

I'm an engineer on the OrcaRouter team.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Minimal coding agent written in Rust, inspired by [pi](https://pi.dev/docs/lates
 
 ## Features
 
-- **Multi-provider**: OpenRouter, OpenAI, Anthropic, Gemini, Ollama, plus custom providers
+- **Multi-provider**: OpenRouter, OrcaRouter, OpenAI, Anthropic, Gemini, Ollama, plus custom providers
 - **Standard tools**: all of the standard tools exposed to coding agents, as described by the opencode documentation.
 - **Permission system**: five configurable modes with per-tool patterns, session allowlists, and configurable mode-to-rule application policies
 - **Session management**: save/load/resume sessions, auto-compaction to stay within context windows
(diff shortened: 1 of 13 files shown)


gi-dellav:
New providers must be a doc entry, like MiniMax, not new source code

PR #270, opened 2026-09-03
Fix CPU=100% after quitting parent terminal

velppa:
Hi Giuseppe, I tried zerostack around v1.3.0, but after it started eating 100% when I quit Terminal.app, I patched it; but never sent a PR. Today I realized that v1.8.0 still suffers from the same problem, so asked Claude to send a PR.

To reproduce the problem, open Terminal.app and zerostack there, then close the Terminal.app and check CPU usage.

_below is AI-generated text_.

## Summary
- `spawn_event_thread`'s poll-error path retried in a tight loop with no backoff once stdin closed (e.g. the parent terminal quit), pinning a CPU core at 100%.
- The poll-error branch now breaks the thread instead of retrying, and sets the shared `running` flag to `false` on every exit path.
- `App::step`'s main select loop now watches `running` on a 250ms tick and shuts the app down once the event thread is gone, instead of idling forever with no input source.

This ports the fix originally landed against v1.3.0 (the pre-refactor `src/ui/mod.rs` event loop) onto the current `app.rs`/`mod.rs` split, where the same unhandled-poll-error pattern was still present.

## Test plan
- [x] `cargo check --all-features`
- [x] `cargo test --all-features ui::` (all passing)
- [ ] Manual: launch, close the parent terminal, confirm the orphaned process exits instead of spinning at 100% CPU

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01WSMoyfrah7srH1CuatH73i

### DIFF
--- a/src/ui/app.rs
+++ b/src/ui/app.rs
@@ -501,6 +501,15 @@ impl<'a> App<'a> {
                 _ = tokio::time::sleep(Duration::from_millis(100)), if self.run.is_running => {
                     self.refresh()?;
                 }
+                _ = tokio::time::sleep(Duration::from_millis(250)) => {
+                    // The event thread sets this to false when it exits
+                    // (e.g. stdin closed because the parent terminal quit).
+                    // Without this check the app would otherwise idle
+                    // forever with no way to receive input.
+                    if !self.running.load(Ordering::Relaxed) {
+                        return Ok(ControlFlow::Break(()));
+                    }
+                }
                 else => {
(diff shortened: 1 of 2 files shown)


gi-dellav:
Hi, nice catch, just merged!

p.s. If you want to keep up with the updates and provide feedback on the project, consider joining our Matrix chatroom https://matrix.to/#/#zerostack-general:matrix.org

PR #281, opened 2026-09-08
fix(ui): make bash allow-always patterns match paths

NotAFlightRisk:
When you press `a` on a bash permission prompt for a command containing `/`, youre asked again on the very next call, because the suggested session rule doesn't match the command it came from. The broken rule is persisted, so the prompts carry on after a resume too.

Since commit 5c1aa4e (v1.8.2), this affects the default Standard mode. A stock install prompts for `sudo cp app.conf /etc/app.conf` under the new `("sudo", Ask)` and `("sudo **", Ask)` rules, but pressing `a` still leaves the next sudo command with a path asking again.

Bare commands remain uncovered: `pytest` suggests `pytest **`, which expects a space. That was already the case, and I couldnt tell whether allowing a bare command should cover only that command or its whole family.

The suggester uses a single `*`, and `glob_to_regex` compiles that as `[^/]*` while `**` becomes `.*`:

| command | suggested | compiles to | matches itself? |
|---|---|---|---|
| `pytest tests/test_login.py` | `pytest *` | `^pytest [^/]*$` | no |
| `npm install ./packages/web` | `npm *` | `^npm [^/]*$` | no |
| `cargo test --all` | `cargo *` | `^cargo [^/]*$` | yes |

Direct denies still win, though compund commands can escape a deny prefix, and that already applies to built-in Allow rules such as `cat **`.

Changing single-star matching would widen user allow and deny patterns, and break the documented `src/*` / `src/**` distinction. The built-in bash rules and every bash example in `docs/CONFIG.md` already use `**`; the suggester was the odd one out.

### DIFF
--- a/src/tests/checker_tests.rs
+++ b/src/tests/checker_tests.rs
@@ -1,5 +1,6 @@
 use crate::permission::checker::{CheckResult, PermissionChecker};
 use crate::permission::{Action, PermissionConfig, PermissionConfigs, SecurityMode, ToolPerm};
+use crate::ui::utils::suggest_pattern;
 
 fn default_modes() -> Option<Vec<String>> {
     Some(vec![
@@ -363,6 +364,14 @@ fn session_allowlist_cannot_bypass_deny_rules() {
         "deny rule must not be bypassed by session allowlist, got {:?}",
         result,
     );
+}
+
+#[test]
+fn session_allowlist_covers_a_bash_command_with_a_path_argument() {
+    let cmd = "pytest tests/test_login.py";
+    let mut checker = make_checker(SecurityMode::Guarded);
+    checker.add_session_allowlist("bash".into(), &suggest_pattern("bash", cmd));
+    assert!(matches!(checker.check("bash", cmd), CheckResult::Allowed));
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


xavierforge (approved):
Thanks for tracking this down, and for the clear write-up with the regex table. Reviewed and verified locally, approving.

- Root cause checks out: `glob_to_regex` compiles `*` to `[^/]*`, so `pytest *` can't match `pytest tests/x.py`. The built-in bash rules and every bash example in docs/CONFIG.md already use `**`; the suggester was the only place producing a single star.
- Reproduced on main in default Standard mode with `sudo -n stat /etc/hostname` followed by `sudo -n stat /etc/os-release`: pressing `a` on the first still prompts on the second. On this branch the second command runs without a prompt.
- The regression test is real: reverting the one-line change in `src/ui/utils.rs` makes it fail.
- Deny rules still take precedence over the session allowlist, so `sudo **` cannot unlock the built-in `sudo rm -rf /**` deny.

The CI failures are not from this PR. The same four jobs (both clippy legs and both `--all-features` test legs) are already red on main at a968535: six clippy errors in `src/engine/mod.rs` and `run_string_reasoning_toggles` panicking with "advisor config not initialized". This branch does not touch either.

Two non-blocking notes:
- Entries already saved to sessions as `foo *` stay broken; pressing `a` once more on the next prompt appends a working `foo **`. A one-off migration in `load_session_allowlist` is possible but probably not worth the code.
- Bare commands (`pytest` with no arguments) still prompt since `^pytest .*$` needs the space. Agreed it's out of scope here.

NotAFlightRisk:
Thanks @xavierforge, and for actually reproducing it rather than taking my word for it. Agreed on both of your notes - I won't chase either.

PR #282, opened 2026-09-09
feat(status-signals): report permission waits (protocol v1.1)

xavierforge:
## Summary

Status-signal protocol v1.1: zerostack now reports the interactive permission wait over `--status-socket` as `blocked:permission` and reports the end of that wait as `state:working`, so an external supervisor can tell "working" apart from "waiting for a human". The three v1.0 lines (`start`, `stop`, `git-conflict`) are byte-for-byte unchanged and a v1.0 consumer that ignores unknown lines sees no difference. A new `docs/STATUS_SIGNALS.md` becomes the single source of truth for the wire protocol.

## Motivation

Tools such as multistack bind the status socket and spawn zerostack to run several agents side by side. Today the last thing they hear before a permission prompt is `start`, so an agent parked on `(y) allow once (a) allow always (n) deny` looks exactly like one that is busy: the panel shows "working", the per-agent timer keeps counting, and nobody gets notified. That is the one moment a parallel-agent manager exists to surface. This change makes it observable with two extra lines and no change to anything a v1.0 listener already parses.

## Design decisions

**Why `state:working` rather than `unblocked:<reason>`.** The release message names the state zerostack is in after the wait, not the inverse of the block. A permission decision resumes the same run, so the only state this version can report is `working`; the reserved `state:idle` exists because a later wait (the `/chain` question, which arrives after `stop`) can end with no run at all when the user declines, and "unblocked" would leave the consumer guessing which of the two it is. The document states that `state:` is level-triggered with no pairing or nesting semantics; waits are serialised today, so a consumer must not build a stack on the pair.

**Why an RAII guard with an explicit `drop`.** `handle_permission_request` has three fallible calls between drawing the prompt and taking the decision. Two bare sends would leave a listener stuck on `blocked` after any `?` early return. `BlockedScope` sends `blocked:permission` on construction and `state:working` on drop, so every exit path, including unwind, stays balanced; the explicit `drop` right before `ask_req.reply.send` is what enforces the documented order (release before the decision reaches the tool), and it is commented as load-bearing. The raw `send_blocked` / `send_state` stay available because the reserved chain wait spans several event-loop turns and cannot be guarded; the doc comments say to prefer the guard.

**Why the transport is untouched.** The sender still connects per message, blocks synchronously, and ignores every error, exactly as v1.0 does. Bounding that (write timeouts, a sender thread) is a real improvement but a separate concern from the protocol extension, and mixing it in would make this diff harder to review. The trust-model section of the protocol document states the consequence plainly: whoever can bind the socket path can observe the timeline and, by never reading, stall the agent; put the socket in a private directory.

**Why `permission_handler` became `pub(crate)`.** The handler-level tests live in `src/tests/` per this repo's convention, and a private module of `ui` is not nameable from there. One visibility keyword, matching every sibling module in `src/ui/mod.rs`, no behaviour change.

**What stays out on purpose.** `blocked:chain`, `blocked:loop-plan`, `state:idle`, a release signal for `git-conflict`, a protocol version banner, and a `ZEROSTACK_STATUS_SOCKET` environment variable are all reserved for later versions; each has its own ordering subtlety and deserves its own change.

## Testing

- Unit tests (`src/tests/status_signals_tests.rs`): exact bytes for the two new lines, guard emits the pair in order, guard releases on a `?` error path, missing socket stays silent for every sender.
- Handler-level tests (`src/tests/permission_signal_tests.rs`): a temp UnixListener plus a fake render backend drive `handle_permission_request` for allow once, allow always, deny, and Esc, asserting exactly `["blocked:permission", "state:working"]` on the wire and the matching decision on the oneshot; a no-socket case asserts nothing is sent; an error-path case uses a backend that fails mid-prompt and asserts the handler returns `Err` while the wire still carries the full pair; every handler call is bounded by a 5-second timeout so a regression fails instead of hanging CI.
- Ordering test: `state_working_reaches_the_wire_before_the_decision` registers a waker on the tool side of the oneshot and snapshots the socket's accept queue from inside `wake_by_ref`, which runs inline in `reply.send`; the snapshot must already contain `state:working`. Moving the guard's drop below the send makes it fail deterministically (checked by mutation); 20 sequential runs, no flake.
- Full suite: `cargo test` 972 passed. `cargo fmt --check` clean. Both clippy legs (`--all-features`, `--no-default-features`) report nothing in the files this branch touches.
- Live verification: the binary built from this branch was driven over a pty against a scripted local provider with a listener on a fresh socket; the observed sequence across one turn with one permission prompt was `start`, `blocked:permission`, `state:working`, `stop`, with the four-second gap between the middle two matching the driver's pause before pressing `y`.
- Line breakdown for reviewers: production +86 / -22 (`status_signals.rs`, `permission_handler.rs`, one keyword in `ui/mod.rs`), tests +571, docs +143 / -9. Most of the test volume is the handler-level harness and the ordering test.
- vibe-diff: not run
- cross-check: not run
- verify: /opsx:verify ran in-session; no CRITICAL findings, one WARNING (leftover `#[allow(dead_code)]`) fixed before shipping.
- adversarial review: opus, fresh context, upstream-maintainer persona; 17 findings, all either folded into the commits above or answered in Design decisions and Notes.

## Reviewing

Start with `docs/STATUS_SIGNALS.md` (the contract), then `src/extras/status_signals.rs` (the guard, about 70 lines), then the eleven-line change in `src/ui/permission_handler.rs`. The test files are mechanical: `status_signals_tests.rs` pins bytes, `permission_signal_tests.rs` pins the bracket per decision and, in its last test, the ordering. `docs/CONFIG.md`, `README.md`, and `docs/index.html` only shrink to a link.

## Notes

- CI is expected to be red on clippy for reasons unrelated to this branch: `main` at a968535 carries six clippy errors in `src/engine/mod.rs` (`too_many_arguments`, `needless_borrows_for_generic_args` x3, `never_loop`, `unnecessary_to_owned`, plus `needless_return` on the `--no-default-features` leg). This branch does not touch that file; filtering it out of either clippy leg leaves zero diagnostics. Cleaning it up belongs in its own `fix(engine)` change.
- The cross-check and vibe-diff review layers were waived by the author for this branch; the in-session verify pass and a fresh-context adversarial review (whose findings are folded in here) were run instead.
- The status line goes blank while the permission prompt is displayed (`permission_handler` passes an empty status line to `draw_bottom`). Pre-existing, unrelated to signals, left for a separate fix.
- Follow-ups reserved by the protocol document: `blocked:chain` (ordering is `stop` then `blocked:chain`, so a consumer sees a brief idle), `blocked:loop-plan`, `state:idle`, and a release signal for `git-conflict`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Minimal coding agent written in Rust, inspired by [pi](https://pi.dev/docs/lates
 - **Subagents**: Parallel and fast, used for exploring the codebase
 - **ARCHITECTURE.md**: Our own companion file for AGENTS.md, it allows to offer a shared core knowledge for all agents working on the same codebase
 - **Prompt chaining**: offers to advance brainstorm → plan → code → review as each phase finishes, config-gated per transition
-- **Status signals**: emits start/stop/git-conflict events over a Unix socket for external status bars or tooling
+- **Status signals**: emits run-state events (start, stop, git-conflict, permission waits) over a Unix socket for external status bars or tooling ([docs/STATUS_SIGNALS.md](docs/STATUS_SIGNALS.md))
 
 **NOTE**: Windows support is not tested is any way, but feel free to try and open an issue if you encounter any bugs!
 
@@ -391,9 +391,9 @@ provider's multimodal support.
 **NOTE:** Status signals require the `status-signals` feature, which is
 included in the default build.
 
-Pass `--status-socket <path>` to have zerostack emit `start`, `stop`, and
-`git-conflict` events over a Unix domain socket, for external status bars or
-tooling to watch.
+Pass `--status-socket <path>` to have zerostack report its run state over a
+Unix domain socket, for external status bars or tooling to watch. See
+[docs/STATUS_SIGNALS.md](docs/STATUS_SIGNALS.md) for the full protocol.
 
 ## Parallel Agent
 
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -763,10 +763,9 @@ when one of them is used). All other items are read from the session.
 ## Status signals
 
 Requires the `status-signals` feature (included in the default build). Pass
-`--status-socket <path>` to have zerostack emit `start`, `stop`, and
-`git-conflict` events over a Unix domain socket at `<path>`, for external
-status bars or tooling to watch. This is separate from the in-TUI status bar
-above.
+`--status-socket <path>` to have zerostack report its run state over a Unix
+domain socket at `<path>`, for external status bars or tooling to watch. See
+[STATUS_SIGNALS.md](STATUS_SIGNALS.md) for the full protocol.
 
 ## Colors
 
--- a/docs/STATUS_SIGNALS.md
+++ b/docs/STATUS_SIGNALS.md
@@ -1,0 +1,135 @@
+---
+description: "Wire protocol for the status-signal Unix domain socket: every message, its ordering guarantees, the run-boundary invariant, listener rules, and which modes emit what."
+---
+
+# Status signals
+
+This is the single source of truth for the status-signal protocol. `docs/CONFIG.md` and
+`README.md` link here instead of restating the message list; if you are editing the
+protocol, this is the only file that needs to change.
+
+## Overview
+
+zerostack can report its run state to an external supervisor (a status bar, a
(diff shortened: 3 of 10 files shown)


gi-dellav:
Great idea!

PR #286, opened 2026-09-12
fix(ci): clear engine clippy lints and the advisor panic under --all-features

xavierforge:
main has been red since a968535 (the Engine module), which landed without a CI run. Four jobs fail: both clippy legs and both `--all-features` test legs. #281 and #282 inherited the same failures.

## Commit 1: `fix(engine): clear clippy lints in the engine module`

Seven lints, all in `src/engine/mod.rs`, no `#[allow]` added:

- `needless_borrows_for_generic_args` on three `write_error` / `write_line` calls (the sink methods take `impl AsRef<str>`).
- `unnecessary_to_owned` in the `/btw` handler.
- `never_loop` in the `/btw` handler: the `while let` over the `BtwEvent` channel breaks or returns on every arm, so it is an `if let`.
- `needless_return` in the `not(feature = "memory")` branch of `slash_memory` (only visible on the `--no-default-features` leg).
- `too_many_arguments` on `Engine::with_agent` (8/7). It becomes a consuming builder method, so tests now write `Engine::new(...).with_agent(agent)`. Three call sites in `src/tests/engine_tests.rs` updated; no other callers.

## Commit 2: `fix(advisor): treat uninitialized config as disabled when building agents`

`build_agent` called `advisor::with_config(|c| c.enabled)` unconditionally, and `with_config` panics with "advisor config not initialized" when `init_config` has not run. Headless engines that rebuild the agent (for example `/reasoning`) reach this before advisor startup; `run_string_reasoning_toggles` hit it under `--all-features`.

New `advisor::is_enabled()` returns `false` when the config is not initialized, and `build_agent` uses it. No test was changed: the failing test now passes because of the production fix and doubles as the regression test for rebuilding an agent without advisor startup.

## Verification

- `cargo fmt --check`
- `cargo clippy --locked --all-targets --all-features -- -D warnings`
- `cargo clippy --locked --all-targets --no-default-features -- -D warnings`
- `cargo test --locked --all-features`: 1194 passed, 0 failed
- `cargo test --locked --no-default-features`: 808 passed, 0 failed

### DIFF
--- a/src/agent/builder.rs
+++ b/src/agent/builder.rs
@@ -372,7 +372,7 @@ pub async fn build_agent_inner<M: CompletionModel + 'static>(
         }
 
         #[cfg(feature = "advisor")]
-        if crate::extras::advisor::with_config(|c| c.enabled) {
+        if crate::extras::advisor::is_enabled() {
             use crate::extras::advisor::AdvisorTool;
             all_tools.push(Box::new(AdvisorTool::new()));
         }
--- a/src/engine/mod.rs
+++ b/src/engine/mod.rs
@@ -97,7 +97,7 @@ impl RunOutput {
 }
 
 /// Owned headless execution context. Construct with [`Engine::new`] (fresh
-/// in-process state, no disk writes unless a command needs them) or
+/// in-process state, no disk writes unless a command needs them), then chain
 /// [`Engine::with_agent`] to inject a scripted agent in tests.
 pub struct Engine {
     cli: Cli,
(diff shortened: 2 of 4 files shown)


xavierforge:
@gi-dellav heads-up before this lands, since it touches the Engine module you pushed in a968535:

- `Engine::with_agent` changes shape. The 8-argument constructor is gone; it is now a consuming builder, so callers write `Engine::new(...).with_agent(agent)`. If you have local engine work still on the old signature it will need that one-line adjustment after merging. If you would rather keep the original constructor with `#[allow(clippy::too_many_arguments)]`, say so and I will switch it, no strong preference here.
- `build_agent` no longer panics when advisor config has not been initialized; `advisor::is_enabled()` treats that as disabled. Startup still calls `init_config` before any `build_agent`, so TUI and headless behaviour are unchanged. Only the rebuild-before-startup path (which `run_string_reasoning_toggles` exercised) changes.

Merge with a merge commit or rebase rather than squash, so the lint commit and the advisor commit stay separate.

gi-dellav:
All perfect, thank you very much for the PR + for the writeup

PR #299, opened 2026-09-21
add darcs commands to permissions

toastal:
I have been using Darcs as my primary VCS for the past 2 years & as such I have a lot of projects that could use the basic commands. I believe these are the basic commands an LLM might need to poke around a project. I excluded the prehook & posthook commands as they allow arbitrary code execution.

NOTE: I have never written Rust so I am just copying the surrounding patterns. `cargo test` seems to pass.


xavierforge:
Thanks for this, the shape is right and it fits the surrounding rules well.
Deny rules are evaluated before everything else in the checker, so the
hook denies work regardless of where they sit in the list, and CI is green
across all legs.

One gap I noticed while reading the darcs man page: `darcs diff` takes
`--diff-command "<program> %1 %2"`, which replaces the built-in diff with
an arbitrary external program (arguments allowed, just no shell syntax).
That is the same class of thing as `--prehook` / `--posthook`, so it
probably deserves the same treatment:

    ("darcs **--diff-command**", Action::Deny),

plus a case in `restrict_darcs_hooks`, something like
`darcs diff --diff-command \"sh -c 'rm -rf /' x %1 %2\"`.

You know darcs a lot better than I do, so a broader question: are there
other flags on `annotate`, `log`, `show`, `test`, or `whatsnew` that can
run a program, write outside the repo, or otherwise escape the read-only
intent? If so it would be great to fold them into the same deny block
while you are here.

For the record, `darcs test` running a `setpref test` command recorded in
the repo history is fine by me. It is the same trust boundary as
`cargo test`, `npm run`, and `pixi run`, which are already allowed by
default.

Happy to merge once the `--diff-command` deny is in.

toastal:
That’s a good catch. I use `darcs diff --diff-command "kitten diff %1 %2"` all the time locally, but not useful for an LLM diffing code for review.

xavierforge:
Looks good, thanks for the quick turnaround. Merging.

